PR #31, opened 2022-03-28
Support ACM 2.5

clyang82:
couple of changes:
1. there is a new component `grc-policy-addon-controller` introduced for grc and remove the `grc-grcui` and `grc-grcuiapi`. so we need to fork it and add back the removed components.
2. update the `governance-policy-propagator` to align with `release-2.5`.
3. Get the `SNAPSHOT` environment to enable deploy ACM Hub with the specified snapshot.

### DIFF
--- a/deploy/deploy_hub_of_hubs.sh
+++ b/deploy/deploy_hub_of_hubs.sh
@@ -71,13 +71,19 @@ function deploy_hoh_controllers() {
     TRANSPORT_TYPE="${transport_type}" IMAGE="quay.io/open-cluster-management-hub-of-hubs/hub-of-hubs-status-transport-bridge:$TAG" envsubst | kubectl apply -f - -n "$acm_namespace"
 
   # deploy hub cluster controller
-  deploy_component "hub-cluster-controller" "$branch" "kubectl apply -n $acm_namespace -k ./deploy"
+  deploy_component "hub-cluster-controller" "$branch" deploy_hub_cluster_controller_action
 
   # deploy hub of hubs addon controller
   deploy_component "hub-of-hubs-addon" "$branch" deploy_hub_of_hubs_addon_action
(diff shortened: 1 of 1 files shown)


deploy/deploy_hub_of_hubs.sh
```diff
@@ -132,8 +132,9 @@ function deploy_grc_chart_action() {
   kubectl delete appsub grc-sub -n "$acm_namespace" --ignore-not-found
 
   cat values.yaml |
-      yq e ".global.imageOverrides.governance_policy_propagator = \"quay.io/open-cluster-management-hub-of-hubs/governance-policy-propagator:no_status_update\"" - |
+      yq e ".global.imageOverrides.governance_policy_propagator = \"quay.io/open-cluster-management-hub-of-hubs/governance-policy-propagator:hub-of-hubs\"" - |
       yq e ".global.imageOverrides.grc_ui = \"quay.io/open-cluster-management-hub-of-hubs/grc-ui:$TAG\"" - |
+      yq e ".global.imageOverrides.grc_ui_api = \"quay.io/stolostron/grc-ui-api:2.4.3-SNAPSHOT-2022-03-21-21-19-21\"" - |
```
nirrozenbaum, line 137: @clyang82 looks good.
one comment/question in line #137 - I see you specified a specific image tag `2.4.3-SNAPSHOT-2022-03-21-21-19-21`. don't we want it to use same tag as other images?

clyang82: in ACM 2.5, `grc-ui-api` was removed. it was mixed into `console` component. I just picked up a latest version for us from official repo, because we do not touch this component. That is why I did not use the same tag for it.

nirrozenbaum (approved):
/lgtm

PR #32, opened 2022-04-08
deploy observability for hoh.

morvencao:

### DIFF
--- a/deploy/deploy_hub_of_hubs.sh
+++ b/deploy/deploy_hub_of_hubs.sh
@@ -14,6 +14,18 @@ if [ $TAG == "latest" ]; then
   branch="main"
 fi
 
+mco_exists=$(kubectl get multiclusterobservabilities --ignore-not-found)
+if [[ -z "$mco_exists" ]]; then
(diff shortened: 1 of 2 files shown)


deploy/README.md
```diff
@@ -30,6 +30,15 @@
...
     ``` 
 
+1.  If you want to enable Observability service for Hub of Hubs, set the observability enable flag variable and the required variables for AWS S3 storage:
+    ```
+    export ENABLE_OBSERVABILITY=true
+    export S3_BUCKET=<YOUR_AWS_S3_BUCKET>
+    export S3_REGION=<YOUR_AWS_S3_REGION>
+    export AWS_ACCESS_KEY_ID=<YOUR_AWS_ACCESS_KEY_ID>
+    export AWS_SECRET_ACCESS_KEY=<YOUR_AWS_SECRET_ACCESS_KEY>
+    ```
```
clyang82, line 40: I do not expect to enable observability during hub of hubs setup. We have the official document to enable observability on ACM. may notify the user if the observability is not enabled. for example: if the observability is not enabled, users can choose to quit the hub of hubs installation to install observability before setting up hub of hubs.

morvencao: fixed.

PR #36, opened 2022-05-01
Update deploy_hub_of_hubs.sh

nirrozenbaum:


clyang82:
Good catch. Thanks @nirrozenbaum 

/lgtm

PR #41, opened 2022-05-27
fix application image patch command.

morvencao:
fix application image patch command, otherwise, we will see the following error:

image


clyang82:
Good catch.

/lgtm

PR #42, opened 2022-05-31
Enable the application.app.k8s.io and route.openshift.io for kind cluster

yanmxa:


morvencao (commented):
/lgtm

PR #60, opened 2022-06-28
Expose kafka with nodeport

yanmxa:
fixes: https://github.com/stolostron/backlog/issues/23767


test/setup/kafka/kafka-cluster.yaml
```diff
@@ -51,31 +54,3 @@ spec:
...
-metadata:
-  name: status
-  namespace: kafka
-  labels:
-    strimzi.io/cluster: kafka-brokers-cluster
-spec:
-  partitions: 1
-  replicas: 2
-  config:
-    cleanup.policy: compact
```
morvencao, line 81: we don't need these two resources in testing any more?

yanmxa: Because it's more about the hoh, not the Kafka itself, I move it to hub-of-hubs setup later.

yanmxa: I discussed with @clyang82 , and we agreed that should keep Postgres and Kafka middleware as separate as possible. Such as the Kafka cluster instance. we wanted i to contain less information about HoH.

test/setup/kind/top-hub-config.yaml
```diff
@@ -0,0 +1,17 @@
+kind: Cluster
```
clyang82, line 1: suggest to use hoh-config.yaml

test/setup/kafka/kafka-cluster.yaml
clyang82: I am not quit sure why remove the topics. the topic is used to organize messages. HoH should have 2 topics.

yanmxa: Just like we move the postgres-init.yaml to hoh_setup.sh. I move the topics which are specific for hoh to hoh_setup.sh too.

morvencao (commented):
/LGTM

PR #61, opened 2022-06-28
update leader election settings for hoh agent.

morvencao:
try to fix: https://github.com/stolostron/backlog/issues/23768


clyang82:
That is ACM standard to use customized values for controller runtime.
I am not sure if the bug was fixed or not. but anyway, we need to have this code.

/lgtm

PR #64, opened 2022-06-30
fix the problem that the postgres cannot be ready in microshift env

clyang82:
the postgres pods cannot be started successfully due to the permission issue. normally, the KinD or Openshift environmet can inject `fsGroup` to make sure the mount point by the correct group. but in microshift, there is not `fsGroup` injected.

Will log a microshift issue soon.


morvencao (commented):
/lgtm

PR #77, opened 2022-07-07
deploy a local policy to verify the functions

yanmxa:
Fixed: https://github.com/stolostron/backlog/issues/23921

### DIFF
--- a/cicd-scripts/run-local-e2e-test.sh
+++ b/cicd-scripts/run-local-e2e-test.sh
@@ -8,6 +8,7 @@ OPTIONS_FILE="${ROOT_DIR}/test/resources/options-local.yaml"
 HUB_OF_HUB_NAME="hub-of-hubs" # the container name
 HUB_OF_HUB_CTX="microshift"
 
+# KinD cluster, the context with prefix 'kind-'
 LEAF_HUB_NAME="hub1"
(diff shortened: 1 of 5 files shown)


morvencao (changes requested):
You need to add label `hub-of-hubs.open-cluster-management.io/local-resource: ""` to the local policy resource to indicate it is a local policy.
Also the verification is insufficient, we should check which cluster is in NonCompliant status, to make sure the policy is applied to the expected leaf hub.
For the local enforce policy, do we need also check the `limitrange` is created in leaf hub?

yanmxa:
> You need to add label `hub-of-hubs.open-cluster-management.io/local-resource: ""` to the local policy resource to indicate it is a local policy. Also the verification is insufficient, we should check which cluster is in NonCompliant status, to make sure the policy is applied to the expected leaf hub. For the local enforce policy, do we need also check the `limitrange` is created in leaf hub?

I added the label to all the local policy resources and created a new namespace for the local policy. After that, we can get the policy compliant and cluster information, which it is scheduled to, from the policy status. 

We don't need to check the `limitrange` CR in the leaf hub. If the compliantState of the policy status is `Compliant`, it means that the `limitrange` CR is already in the leaf hub. So we just need to check the `end` result.

morvencao (commented):
/LGTM

PR #87, opened 2022-07-13
fix cannot propogate local policy

clyang82:
`HUB_KUBECONFIG` variable was overwritten in `leafhub_setup.sh`. so the `initPolicy` is using leaf hub kubeconfig. that is why cannot watch the local policy.
Replace with `HOH_KUBECONFIG`


morvencao (commented):
/lgtm

PR #90, opened 2022-07-18
Reduce the resources usage

clyang82:
1. remove finalizer. it can be added automatically
2. set `klusterlet` operator replicas to 1
3. set `zookeeper` replicas to 1
4. only install route CRD for pacman application
5. set `hub-of-hubs-rbac` replicas to 1

### DIFF
--- a/test/resources/app/app-pacman-appsub.yaml
+++ b/test/resources/app/app-pacman-appsub.yaml
@@ -22,8 +22,6 @@ kind: Application
 metadata:
   annotations:
     apps.open-cluster-management.io/deployables: ''
-  finalizers:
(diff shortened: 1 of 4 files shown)


clyang82:
kafka broker throws this error.
```2022-07-18 03:32:50,762 INFO [Admin Manager on Broker 0]: Error processing create topic request CreatableTopic(name='status', numPartitions=1, replicationFactor=2, assignments=[], configs=[CreateableTopicConfig(name='cleanup.policy', value='compact')]) (kafka.server.AdminManager) [data-plane-kafka-request-handler-6]```
so need to keep broker with 3 replicas.

PR #91, opened 2022-07-18
add agent resources applier.

morvencao:
reuse hub-of-hubs agent image to create CRDs deploy job in hypershift hosting cluster.


clyang82:
You may need to remove the manifests in pkg/applier since you are using the ones under pkg/cmd.

agent/cmd/applier/main.go
```diff
@@ -0,0 +1,82 @@
...
+
+	// create dynamic client
+	dyn, err := dynamic.NewForConfig(cfg)
+	if err != nil {
+		log.Error(err, "failed to create dynamic client")
+		return 1
+	}
+
+	// apply the manifests
+	if err := applier.ApplyManifestsForVersion(context.TODO(), hubVersion, dyn, mapper, log); err != nil {
```
clyang82, line 65: We need to a switch to trigger it. it is only for hypershift hosted cluster.

morvencao: trigger is at hub-cluster-controller, job for this will be created when hypershift hosted cluster is created/updated.(cover upgrade case)

clyang82:
I am fine to maintain the crds in hub of hubs agent temporarily for release 2.6 dev preview and apply them by a job. for GA, we need to find another alternative, because we should not maintain crds, it should be pushed from somewhere. the job is not a reconcile way.

BTW, I think for 2.6, we just need 2 dirs. One is 2.5 and the other is 2.6. What do you think?  2.6 is for `latest` image testing, while 2.5 covers all 2.5 z stream release.

morvencao:
another look? @clyang82

PR #110, opened 2022-08-08
fix operator aggregated role.

morvencao:
ref: https://github.com/stolostron/backlog/issues/24609


clyang82:
/lgtm

We may need to find a way in the future to ensure the permission is updated in manager can be updated in operator as well. maybe diff 2 files?

morvencao:
we should have script to sync the manager roles/clusteroles to the operator manifests.

PR #112, opened 2022-08-09
Provide ability to create postgres and kafka for operator local testing

yanmxa:
Fixed: https://github.com/stolostron/backlog/issues/24951


operator/README.md
```diff
@@ -25,6 +21,11 @@ kubectl create secret generic kafka-secret -n "open-cluster-management" \
     --from-literal=CA=<CA-for-kafka-server>
 ```
 
+5. You can run the following script to create kafka and postgres clusters and the corresponding secrets in a local environment (the names of the two secrets should be strictly the same as the names of postgres and kafka in the hub of hubs CR)
+    - Install OLM(Optional): `./operator/config/samples/olm.sh`
```
clyang82, line 25: We do not need to install OLM I guessed, because the hub of hubs cluster is an OCP cluster. there is OLM installed. right?

operator/README.md
```diff
@@ -10,11 +10,7 @@ The operator of Hub-of-Hubs (see: https://github.com/stolostron/hub-of-hubs)
 
 ```bash
 kubectl create secret generic postgresql-secret -n "open-cluster-management" \
-    --from-literal=host=<postgresql-host> \
-    --from-literal=port=<postgresql-port> \
-    --from-literal=user=<admin-user-of-hub-of-hubs-database> \
-    --from-literal=password=<admin-user-password> \
-    --from-literal=database=<name-of-hub-of-hubs-database>
+    --from-literal=database_uri=<postgresql-uri>
```
clyang82, line 13: Could you show an example format for `postgresql-uri`?

operator/README.md
```diff
@@ -25,6 +21,11 @@ kubectl create secret generic kafka-secret -n "open-cluster-management" \
     --from-literal=CA=<CA-for-kafka-server>
 ```
 
+5. You can run the following script to create kafka and postgres clusters and the corresponding secrets in a local environment (the names of the two secrets should be strictly the same as the names of postgres and kafka in the hub of hubs CR)
```
clyang82, line 24: suggest to replace `in a local environment` with `in cluster`

operator/README.md
```diff
@@ -25,6 +21,11 @@ kubectl create secret generic kafka-secret -n "open-cluster-management" \
...
+    - Install OLM(Optional): `./operator/config/samples/olm.sh`
+    - Install Kafka and create secret `kafka-secret`: `./operator/config/samples/kafka/deploy_kafka.sh`
```
clyang82, line 26: How about append it to bullet 3. for example:
> PostgreSQL is installed and a database is created for hub-of-hubs, also a secret with name `postgresql-secret` that contains the credential should be created in `open-cluster-management` namespace:

```bash
kubectl create secret generic postgresql-secret -n "open-cluster-management" \
    --from-literal=database_uri=<postgresql-uri>

You can run this sample script - ./operator/config/samples/kafka/deploy_kafka.sh to install kafka in kafka namespace and create the secret automatically.

same as postgreSQL. WDYT?

File: operator/pkg/controllers/hubofhubs/database.go [hunk 5]
```diff
@@ -41,20 +38,14 @@ func (reconciler *ConfigReconciler) reconcileDatabase(ctx context.Context, confi
...
-	conn, err := pgx.Connect(ctx, adminDatabaseURI)
+	databaseURI := string(postgreSecret.Data["database_uri"])
+	conn, err := pgx.Connect(ctx, databaseURI)
 	if err != nil {
 		return fmt.Errorf("failed to connect to postgres: %w", err)
 	}
 	defer conn.Close(ctx)
 
-	err = iofs.WalkDir(databaseFS, "database", func(file string, d iofs.DirEntry, beforeError error) error {
+	err = iofs.WalkDir(databaseFS, "database/v2", func(file string, d iofs.DirEntry, beforeError error) error {
```
clyang82, line 48: what is different between v1 and v2?

clyang82 (changes requested):

operator/README.md
yanmxa: Yeah, Sometimes it runs in OCP, but sometimes it runs in a local cluster like KinD.

operator/pkg/controllers/hubofhubs/database.go
yanmxa: The v1 scripts will create the transport user and process user for hub of hubs to access the database. But the scripts in v2 dictionary will not create users and only use the `postgresql-uri` to access the hub of hubs.

operator/README.md
clyang82: this document is used for the customers. we do not support to run hub of hubs in non-OCP cluster. for the KinD cluster, that is only for testing. install OLM should be part of that test. no here. so we need to remove olm.sh in this folder.

operator/pkg/controllers/hubofhubs/manifests/manager/deployment.yaml
```diff
@@ -37,19 +37,19 @@ spec:
...
-                  key: process_uri
+                  name: "{{.DBSecret}}"
+                  key: database_uri
             - name: TRANSPORT_BRIDGE_DATABASE_URL
               valueFrom:
                 secretKeyRef:
-                  name: {{.DBSecret}}
-                  key: transport_uri
+                  name: "{{.DBSecret}}"
+                  key: database_uri
```
morvencao, line 52: Looks like we can use only one env and pass the env value to different parameters.

clyang82:
💯 
/lgtm

PR #113, opened 2022-08-09
Add default placement

clyang82:
fixed: https://github.com/stolostron/backlog/issues/24615

### DIFF
--- a/operator/pkg/controllers/hubofhubs/config_controller.go
+++ b/operator/pkg/controllers/hubofhubs/config_controller.go
@@ -297,7 +297,8 @@ func (r *ConfigReconciler) manipulateObj(ctx context.Context, objs []*unstructur
 
 func (r *ConfigReconciler) initFinalization(ctx context.Context, hohConfig *hubofhubsv1alpha1.Config,
 	log logr.Logger) (bool, error) {
-	if hohConfig.GetDeletionTimestamp() != nil && utils.Contains(hohConfig.GetFinalizers(), constants.HoHOperatorFinalizer) {
+	if hohConfig.GetDeletionTimestamp() != nil &&
+		utils.Contains(hohConfig.GetFinalizers(), constants.HoHOperatorFinalizer) {
 		log.Info("to delete hoh resources")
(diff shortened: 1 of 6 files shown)


operator/pkg/controllers/hubofhubs/config_controller.go
```diff
@@ -282,6 +262,39 @@ func (r *ConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+		labels := obj.GetLabels()
+		if labels == nil {
+			labels = make(map[string]string)
+		}
+		labels[constants.HoHOperatorOwnerLabelKey] = hohConfig.GetName()
+		obj.SetLabels(labels)
+
+		log.Info("Creating or updating object", "object", obj)
+		if err := hohDeployer.Deploy(obj); err != nil {
+			conditionError := condition.SetConditionRBACDeployed(ctx, r.Client, hohConfig, condition.CONDITION_STATUS_FALSE)
```
morvencao, line 288: This is not always set RBAC deploying condition.
We need to get the current deploying component and then set correct condition for condition.

operator/pkg/controllers/hubofhubs/manifests/placement/placement.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: cluster.open-cluster-management.io/v1beta1
+kind: Placement
+metadata:
+  name: default
+  namespace: open-cluster-management
+  labels:
+    hub-of-hubs.open-cluster-management.io/used-by: hub-of-hubs
```
morvencao, line 7: Is there special meaning of this label?
If not, I would prefer to remove label in each manifests, we should add universal label in operator deployer logic, what do you think?

morvencao (changes requested):

operator/pkg/controllers/hubofhubs/manifests/placement/placement.yaml
clyang82: this is a special one which is used by console. the customers can create their owned placement with this label so that it can be listed in creation managed cluster stage.

PR #115, opened 2022-08-09
rename hoh operator

morvencao:
ref: https://github.com/stolostron/backlog/issues/24609


operator/config/crd/bases/hubofhubs.open-cluster-management.io_multiclusterglobalhubs.yaml
```diff
@@ -5,20 +5,21 @@ metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.8.0
   creationTimestamp: null
-  name: configs.hubofhubs.open-cluster-management.io
+  name: multiclusterglobalhubs.hubofhubs.open-cluster-management.io
 spec:
   group: hubofhubs.open-cluster-management.io
```
clyang82, line 10: suggest to use `operator.open-cluster-management.io` as group name.

clyang82 (changes requested):

clyang82:
fix the code smell. most of them are `which is greater than 120 authorized`. I am looking for a tool to configure in vscode to avoid such error.

operator/config/crd/bases/hubofhubs.open-cluster-management.io_multiclusterglobalhubs.yaml
morvencao: done

PR #118, opened 2022-08-10
Fixed the samples kafka and postgre issues

yanmxa:

### DIFF
--- a/operator/config/manager/kustomization.yaml
+++ b/operator/config/manager/kustomization.yaml
@@ -1,5 +1,5 @@
 resources:
-- manager.yaml
+  - manager.yaml
 
 generatorOptions:
   disableNameSuffixHash: true
@@ -12,6 +12,6 @@ generatorOptions:
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
-- name: quay.io/stolostron/hub-of-hubs-operator
-  newName: quay.io/clyang82/hub-of-hubs-operator
-  newTag: latest
+  - name: quay.io/stolostron/hub-of-hubs-operator
+    newName: quay.io/stolostron/hub-of-hubs-operator
+    newTag: latest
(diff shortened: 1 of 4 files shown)


test/setup/postgres/postgres_setup.sh
```diff
@@ -4,9 +4,8 @@ KUBECONFIG=${1:-$KUBECONFIG}
```
clyang82, line 4: can we reuse `https://github.com/stolostron/hub-of-hubs/tree/main/operator/config/samples/postgres`? same as kafka.

yanmxa: I think it's better to separate the two, because the `sample/postgres` is mainly for OCP environment, while the `setup/postgres` is mainly for local VM installation, There are still some differences between both. For example, in the microshift environment, some file permissions need to be set, but OCP does not, as shown below:
```
kubectl patch statefulset ${sts} -n $pgnamespace -p '{"spec":{"template":{"spec":{"securityContext":{"fsGroup":26}}}}}'
```

yanmxa:
Resolved: https://github.com/stolostron/backlog/issues/24610

test/setup/postgres/postgres_setup.sh
clyang82: why cannot we use `sample/postgres` and then patch for non-OCP environment?

operator/config/manager/kustomization.yaml
```diff
@@ -13,5 +13,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: quay.io/stolostron/hub-of-hubs-operator
-  newName: quay.io/stolostron/hub-of-hubs-operator
+  newName: quay.io/clyang82/hub-of-hubs-operator
```
morvencao, line 16: is this intent?

morvencao (commented):
/lgtm

PR #122, opened 2022-08-11
add support for hub-of-hubs operator to watch owned resources

morvencao:
ref: https://github.com/stolostron/backlog/issues/25023

According to https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/predicate/predicate.go#L140
We should use `GenerationChangedPredicate` if we only want to trigger reconcile when spec changes. @clyang82 

### DIFF
--- a/operator/main.go
+++ b/operator/main.go
@@ -22,6 +22,13 @@ import (
 
 	operatorsv1 "github.com/operator-framework/operator-lifecycle-manager/pkg/package-server/apis/operators/v1"
 	hypershiftdeploymentv1alpha1 "github.com/stolostron/hypershift-deployment-controller/api/v1alpha1"
+	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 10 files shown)


operator/main.go
```diff
@@ -75,13 +86,77 @@ func main() {
...
+				Label: labels.SelectorFromSet(labels.Set{constants.HoHOperatorOwnerLabelKey: constants.HoHOperatorOwnerLabelVal}),
+			},
+			&rbacv1.ClusterRole{}: {
+				Label: labels.SelectorFromSet(labels.Set{constants.HoHOperatorOwnerLabelKey: constants.HoHOperatorOwnerLabelVal}),
+			},
+			&rbacv1.ClusterRoleBinding{}: {
+				Label: labels.SelectorFromSet(labels.Set{constants.HoHOperatorOwnerLabelKey: constants.HoHOperatorOwnerLabelVal}),
+			},
+			&networkingv1.Ingress{}: {
+				Label: labels.SelectorFromSet(labels.Set{constants.HoHOperatorOwnerLabelKey: constants.HoHOperatorOwnerLabelVal}),
```
clyang82, line 123: Do we create `ingress`?

operator/main.go
```diff
@@ -75,13 +86,77 @@ func main() {
...
+			},
+			&networkingv1.NetworkPolicy{}: {
+				Label: labels.SelectorFromSet(labels.Set{constants.HoHOperatorOwnerLabelKey: constants.HoHOperatorOwnerLabelVal}),
```
clyang82, line 126: Where do we use `NetworkPolicy`?

operator/main.go
```diff
@@ -75,13 +86,77 @@ func main() {
...
+			},
+			&clusterv1beta1.ManagedClusterSetBinding{}: {
+				Label: labels.SelectorFromSet(labels.Set{constants.HoHOperatorOwnerLabelKey: constants.HoHOperatorOwnerLabelVal}),
+			},
+			&clusterv1beta1.Placement{}: {
+				Label: labels.SelectorFromSet(labels.Set{constants.HoHOperatorOwnerLabelKey: constants.HoHOperatorOwnerLabelVal}),
+			},
+			&clusterv1beta1.Placement{}: {
+				Label: labels.SelectorFromSet(labels.Set{constants.HoHOperatorOwnerLabelKey: constants.HoHOperatorOwnerLabelVal}),
```
clyang82, line 135: Duplicated with the above?

clyang82 (changes requested):

operator/pkg/controllers/hubofhubs/multiclusterglobalhub_controller.go
```diff
@@ -594,15 +598,39 @@ func (r *MultiClusterGlobalHubReconciler) SetupWithManager(mgr ctrl.Manager) err
 			return true
 		},
 		UpdateFunc: func(e event.UpdateEvent) bool {
-			return e.ObjectOld.GetResourceVersion() != e.ObjectNew.GetResourceVersion()
+			return e.ObjectOld.GetGeneration() != e.ObjectNew.GetGeneration()
```
clyang82, line 601: I think we can check the spec only for mgh cr, because the mgh cr is created by customer and we only care about the `spec`. for other resources, I think we care about all instead of just for `spec`. for example, if the user remove the label we added, we need to add that label back. so for other resources, use `GetResourceVersion`. Does it make sense?

operator/main.go
morvencao: Yes, non-k8s-apiserver needs this.

operator/main.go
morvencao: hub-of-hubs rbac needs this.

operator/main.go
morvencao: removed duplicated one.

operator/pkg/controllers/hubofhubs/multiclusterglobalhub_controller.go
morvencao: it makes sense.

morvencao:
another look? @clyang82

PR #128, opened 2022-08-15
Share kafka and databse setup between operater samples and e2e setup

yanmxa:
Reference: https://github.com/stolostron/backlog/issues/24610


clyang82:
Don't we need to remove the existing files in `https://github.com/stolostron/multicluster-globalhub/tree/main/test/setup/postgres`?

yanmxa:
> Don't we need to remove the existing files in `https://github.com/stolostron/multicluster-globalhub/tree/main/test/setup/postgres`?

We will remove the `test/setup/postgres` and `test/setup/kafka` files at another PR which will switch the setup process with **operator**.

PR #145, opened 2022-08-17
Labels/Annotations/Finalizers

clyang82:
ref: https://github.com/stolostron/backlog/issues/25025

### DIFF
--- a/doc/reserved_names/README.md
+++ b/doc/reserved_names/README.md
@@ -1,0 +1,27 @@
+# Annotations
+
+List all annotations are used by multicluster global hub.
+
+Annotation | Description
+--- | ----------
+global-hub.open-cluster-management.io/managed-by | This annotation is used to identify the resource is added by global hub operator and global hub agent. Use it to clean up the resource during uninstallation.
+global-hub.open-cluster-management.io/origin-ownerreference-uid |
+global-hub.open-cluster-management.io/local-resource |
+global-hub.open-cluster-management.io/managed-by-hoh |
+global-hub.open-cluster-management.io/skip-console-install |
+
+# Finalizer
+
+Annotation | Description
+--- | ----------
+global-hub.open-cluster-management.io/resource-cleanup |
+
+
+# Labels
+
+Label | Description
+--- | ----------
+global-hub.open-cluster-management.io/used-by: global-hub |
+global-hub.open-cluster-management.io/managed-by |
+global-hub.open-cluster-management.io/hub-install |
+global-hub.open-cluster-management.io/global-agent |
--- a/doc/reserved_names/README.md
+++ b/doc/reserved_names/README.md
@@ -4,10 +4,8 @@ List all labels are used by the multicluster global hub.
 
 Label | Description
 --- | ----------
-global-hub.open-cluster-management.io/managed-by=multicluster-global-hub-operator | This label is applied to the resources which are created by the global hub operator. The global hub operator watches the resources based on this label.
-global-hub.open-cluster-management.io/managed-by=multicluster-global-hub-agent | This label is applied to the resources which are created by the global hub agent. The global hub agent watches the resources based on this label.
-global-hub.open-cluster-management.io/regional-hub=disable | This label is applied to the managedcluster resource. Identify this managed cluster won't have ACM installed by the global hub.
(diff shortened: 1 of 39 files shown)


morvencao (commented):
/lgtm
thanks

doc/reserved_names/README.md
```diff
@@ -0,0 +1,28 @@
...
+global-hub.open-cluster-management.io/global-hub-agent=disable | This label is applied to the managedcluster resource. Identify this managed cluster won't be managed by the global hub.
+global-hub.open-cluster-management.io/local-resource= | This label is added during creating some resources. It is used to identify the resource is only applied to global hub cluster. It won't be transfered to the regional hub clusters.
+
+# Annotations
+
+List all annotations are used by multicluster global hub.
+
+Annotation | Description
+--- | ----------
+global-hub.open-cluster-management.io/managed-by= | This annotation is used to identify the managed cluster is managed by which regional hub cluster.
```
qiujian16, line 19: why do we need both managed-by label/annotation

doc/reserved_names/README.md
```diff
@@ -0,0 +1,28 @@
...
+global-hub.open-cluster-management.io/origin-ownerreference-uid= | This annotation is used to identify the resource is from the global hub cluster. The global hub agent is only handled with the resource which has this annotation.
```
qiujian16, line 20: not quite clear the purpose of this

doc/reserved_names/README.md
```diff
@@ -0,0 +1,28 @@
+# Labels
+
+List all labels are used by the multicluster global hub.
+
+Label | Description
+--- | ----------
+global-hub.open-cluster-management.io/managed-by=multicluster-global-hub-operator | This label is applied to the resources which are created by the global hub operator. The global hub operator watches the resources based on this label.
+global-hub.open-cluster-management.io/managed-by=multicluster-global-hub-agent | This label is applied to the resources which are created by the global hub agent. The global hub agent watches the resources based on this label.
+global-hub.open-cluster-management.io/regional-hub=disable | This label is applied to the managedcluster resource. Identify this managed cluster won't have ACM installed by the global hub.
```
qiujian16, line 9: this one and global-hub-agent seems related. you would not have global-hub.open-cluster-management.io/regional-hub=disable and global-hub.open-cluster-management.io/global-hub-agent=enable, right?

I think you could use one label global-hub.open-cluster-management.io/regional-hub-type, and have different types like "do not install hub", "install unmanaged hub", and "install managed hub"

doc/reserved_names/README.md
clyang82: label is used to watch the resources. it can be supported natively by using `labelselector`.
annotation is used to identify the regional hub cluster in the managed cluster. this annotation is only available in postgreSQL now. That is original design. maybe consider to use `label` in future.

doc/reserved_names/README.md
clyang82: for example: if there is a policy propagates from the global hub cluster to the regional hub cluster. it is used to distinguish the policy is from the global hub cluster with the local policy (existing in the regional hub cluster).

doc/reserved_names/README.md
clyang82: That's a good idea. Let me update it.

doc/reserved_names/README.md
qiujian16: I am still not clear why you need uid. will you use uid to find any other resources? uid itself seems not sufficient to do so.

clyang82: we do not use uid to find any other resources. just use this annotation to distinguish the global policy with local policy. You know that the global policy and local policy are handled separately. they are in different db table.

doc/reserved_names/README.md
clyang82: I have updated to use `NoHubInstall|NoAgentInstall`.
`NoHubInstall` = no hub installed. but the agent is installed.
`NoAgentInstall` = no hub and agent installed. we do not have case to install Hub without agent.

clyang82: maybe use `NoHubAgentInstall` instead. Leave `NoAgentInstall` for future using.

morvencao (commented):
/lgtm

PR #155, opened 2022-08-19
set mgh-images-config with image tags for local testing

yanmxa:


morvencao (commented):
nit: I would prefer to cut the image by the character `@` and `:`
instead of local test or prow test.
I mean in local test, we want also specify image digest, right?
so let's check if the image string contain `@`, then use digest, otherwise use tag.

morvencao (commented):
/lgtm

PR #159, opened 2022-08-19
delete mgh CR resources from manager

yanmxa:
Fixed: https://github.com/stolostron/backlog/issues/25146


operator/config/rbac/role.yaml
```diff
@@ -118,8 +118,12 @@ rules:
   resources:
   - managedclusters
   verbs:
+  - create
+  - delete
```
clyang82, line 122: why need to give permission to create/delete managed cluster to the operator?

operator/pkg/controllers/hubofhubs/multiclusterglobalhub_controller.go
```diff
@@ -631,6 +630,46 @@ func (r *MultiClusterGlobalHubReconciler) pruneHoHResources(ctx context.Context)
...
+		}),
+	}
+	// cleanup open-cluster-management ManagedClusterSetBinding
+	managedclustersetbindings := &clusterv1beta1.ManagedClusterSetBindingList{}
+	if err := r.Client.List(ctx, managedclustersetbindings, listOpts...); err != nil {
+		return err
+	}
+	for idx := range managedclustersetbindings.Items {
+		managedclustersetbinding := &managedclustersetbindings.Items[idx]
+		finalizers := managedclustersetbinding.GetFinalizers()
```
clyang82, line 645: I am not sure if it works or not, because the finalizer is added by manager. so it may have conflicted. that means the manager may be re-adding this finalizer while the operator is deleting the finalizer.

operator/config/rbac/role.yaml
morvencao: Also you can't edit this file, it is automatically generate when run `make manifests`.

go.mod
```diff
@@ -94,6 +94,7 @@ require (
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.2 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
+	github.com/open-cluster-management/api v0.0.0-20210527013639-a6845f2ebcb1
```
morvencao, line 97: duplicated with `open-cluster-management.io/api`? OCM api import path should be `open-cluster-management.io/api`, maybe we have code using wrong API import path.

operator/pkg/controllers/hubofhubs/multiclusterglobalhub_controller.go
```diff
@@ -631,6 +630,46 @@ func (r *MultiClusterGlobalHubReconciler) pruneHoHResources(ctx context.Context)
...
+		if utils.Contains(finalizers, commonconstants.GlobalHubCleanupFinalizer) {
+			newFinalizers := utils.Remove(finalizers, commonconstants.GlobalHubCleanupFinalizer)
+			managedclustersetbinding.SetFinalizers(newFinalizers)
+			err := r.Client.Update(ctx, managedclustersetbinding)
```
morvencao, line 649: you'd better set resourceVersion before updating it, otherwise the apiserver may complain resource conflict.

operator/pkg/controllers/hubofhubs/multiclusterglobalhub_controller.go
```diff
@@ -631,6 +630,46 @@ func (r *MultiClusterGlobalHubReconciler) pruneHoHResources(ctx context.Context)
...
+	if err := r.Client.List(ctx, placements, listOpts...); err != nil {
+		return err
+	}
+	for idx := range placements.Items {
+		placement := &placements.Items[idx]
+		finalizers := placement.GetFinalizers()
+		if utils.Contains(finalizers, commonconstants.GlobalHubCleanupFinalizer) {
+			newFinalizers := utils.Remove(finalizers, commonconstants.GlobalHubCleanupFinalizer)
+			placement.SetFinalizers(newFinalizers)
+			err := r.Client.Update(ctx, placement)
```
morvencao, line 666: Also set resourceVersion to the new placement be the resourceVersion of current existing placement, to avoid resource conflict.

operator/pkg/controllers/hubofhubs/multiclusterglobalhub_controller.go
yanmxa: It works, but as we discussed I have moved this to the manager `prestop`.

manager/Dockerfile
```diff
@@ -18,6 +18,9 @@ ENV USER_NAME=manager
 
 # install operator binary
 COPY --from=builder /workspace/bin/manager /usr/local/bin/manager
+COPY ./manager/scripts/prestop.sh /usr/local/bin/prestop.sh
+RUN curl -LO http://stedolan.github.io/jq/download/linux64/jq && \
+    chmod 755 jq && mv jq /usr/local/bin/jq
```
morvencao, line 23: better to add checksum for the jq, I encountered an issue that yq was downloaded wrong one, but the image build doesn't report that.

manager/scripts/prestop.sh
```diff
@@ -0,0 +1,52 @@
...
+
+LABEL=global-hub.open-cluster-management.io%2Fmanaged-by%3Dmulticluster-global-hub-operator
+
+# ManagedClusterSetBinding
+resourcesLink=apis/cluster.open-cluster-management.io/v1beta1/namespaces/${NAMESPACE}/managedclustersetbindings
+resourceNames=$(curl --cacert ${CACERT} --header "Authorization: Bearer ${TOKEN}" ${APISERVER}/${resourcesLink}?labelSelector=$LABEL | jq .items[].metadata.name)
+for resource in $resourceNames; do
+  resource=$(echo $resource |sed 's/\"//g')
+  resourceLink=$resourcesLink/$resource
+  curl ${APISERVER}/${resourceLink} --cacert ${CACERT} --header "Authorization: Bearer ${TOKEN}" \
```
morvencao, line 19: try this:
```
curl --get \
    --data-urlencode "p1=value 1" \
    --data-urlencode "p2=value 2" \
    http://example.com
    # http://example.com?p1=value%201&p2=value%202
```
and the `LABEL` can be more readable.

manager/scripts/prestop.sh
```diff
@@ -0,0 +1,52 @@
...
+    -X PATCH -H 'Content-Type: application/merge-patch+json' \
+    -d '{
+      "metadata": {
+        "finalizers": []
```
morvencao, line 23: not sure if the will introduce other issue, because you cleaned all the finalizers, I thought we should only clean the finalizer that manager added.

manager/scripts/prestop.sh
```diff
@@ -0,0 +1,52 @@
...
+      }
+    }' | jq .
+echo "pathed resource: $resourceLink"
+done
+
+# ConfigMap
```
morvencao, line 45: your want to delete `multicluster-global-hub-config` in manager?

manager/Dockerfile
yanmxa: Thanks for your information!

manager/scripts/prestop.sh
yanmxa: It seems the configmap is removed after the mgh is deleted in the latest version. So I have removed the code from manager `prestop`

manager/scripts/prestop.sh
yanmxa: I have updated the code to remove the specific finalizer more than cleanup the finalizers.

morvencao (commented):
/lgtm

morvencao (commented):
/lgtm

PR #161, opened 2022-08-22
Init resources

clyang82:
ref: https://github.com/stolostron/backlog/issues/25232

2 fixes:
1. filter out the `local-cluster` to be selected as a regional hub cluster
2. init `clusterrole` and `clusterrolebinding` by agent. the previous way cannot be executed. use kubeclient to create such resources.


morvencao (commented):
/lgtm

PR #163, opened 2022-08-22
use the default control-runtime default values

clyang82:
resolve: https://github.com/stolostron/backlog/issues/24752

the root cause is if there is an existing lease in your environment, it needs spend 137s to acquire lease.
if that is brand new environment, there is no such problem.
we have not decided to use OpenShift default values yet, because it impact on the high available https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#high-availability

so for our case, let me use the default value in this release.
we may expose those values in mgh CRD and pass them into each operator.

I am not sure if it impacts on this issue or not - https://github.com/stolostron/backlog/issues/23768. Let me observe it.

### DIFF
--- a/agent/cmd/agent/main.go
+++ b/agent/cmd/agent/main.go
@@ -6,7 +6,6 @@ import (
 	"os"
 	"runtime"
 	"strconv"
-	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


morvencao (commented):
/lgtm

PR #168, opened 2022-08-24
delete mgh CR resource from agent

yanmxa:
Fixed: https://github.com/stolostron/backlog/issues/25172

### DIFF
--- a/operator/pkg/controllers/hubofhubs/manifests/placement/managedclustersetbinding.yaml
+++ b/operator/pkg/controllers/hubofhubs/manifests/placement/managedclustersetbinding.yaml
@@ -3,5 +3,7 @@ kind: ManagedClusterSetBinding
 metadata:
   name: default
   namespace: {{.Namespace}}
+  labels:
+    global-hub.open-cluster-management.io/local-resource: ""
 spec:
   clusterSet: default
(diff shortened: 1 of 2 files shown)


clyang82:
You also need to add this label for the managedclustersetbinding in the same directory.

yanmxa:
> You also need to add this label for the managedclustersetbinding in the same directory.

Thanks, I didn't find the `managedclustersetbinding` in the agent cluster at the latest time. Since the `managedclustersetbinding` is not needed by the agent cluster, It's better to add the local label.

clyang82:
> > You also need to add this label for the managedclustersetbinding in the same directory.
> 
> Thanks, I didn't find the `managedclustersetbinding` in the agent cluster when I removed the mgh CR. Since the `managedclustersetbinding` is not needed by the agent cluster, It's better to add the local label.

if you check this controller, it should be propagated to the regional hub cluster - https://github.com/stolostron/multicluster-global-hub/blob/main/manager/pkg/specsyncer/spec2db/controller/managedclustersetbinding_spec_sync.go

yanmxa:
> > > You also need to add this label for the managedclustersetbinding in the same directory.
> > 
> > 
> > Thanks, I didn't find the `managedclustersetbinding` in the agent cluster when I removed the mgh CR. Since the `managedclustersetbinding` is not needed by the agent cluster, It's better to add the local label.
> 
> if you check this controller, it should be propagated to the regional hub cluster - https://github.com/stolostron/multicluster-global-hub/blob/main/manager/pkg/specsyncer/spec2db/controller/managedclustersetbinding_spec_sync.go

Thanks!

PR #172, opened 2022-08-26
Remove finalizer in agent

clyang82:
it is unnecessary to add finalizer to the resources in the regional hub cluster.


morvencao (commented):
/lgtm

PR #177, opened 2022-08-26
remove console support

clyang82:
Remove console support in 2.7 since the current console is a prototype. in 2.7 release, we will have API or UI for global hub.


morvencao:
@clyang82 do we want to remove this now? then how we check the functions from console?

morvencao (commented):
@clyang82 do we want to remove this now? then how we check the functions from console?

clyang82:
we should develop the APIs to be able to use by the end user. the current UI is a prototype. it is not for production.

morvencao (commented):
/lgtm

morvencao (commented):
/lgtm

PR #183, opened 2022-08-30
prune policy finalizer from global cluster

yanmxa:
Fixed: https://github.com/stolostron/backlog/issues/25316

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ unit-tests-agent: setup_envtest
 e2e-setup-dependencies: 
 	./test/setup/e2e_dependencies.sh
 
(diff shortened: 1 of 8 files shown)


clyang82 (changes requested):
We still need to handle the following resources:
application
managedclusterset
managedclustersetbinding
placement

yanmxa:
> We still need to handle the following resources: application managedclusterset managedclustersetbinding placement


> managedclusterset
Thanks! I missed the application. But for managedclusterset, managedclustersetbinding and placement, I didn't find the finalizer marked on them. I still deal with these in the latest commit.

PR #189, opened 2022-09-01
Skip application tests temporarily

clyang82:
ref: https://github.com/open-cluster-management-io/multicloud-operators-subscription/issues/247
Will revert this change once this issue is fixed in application side.


morvencao (commented):
/lgtm

PR #193, opened 2022-09-02
Update the api to v1alpha2

clyang82:


operator/pkg/controllers/hubofhubs/multiclusterglobalhub_controller.go
```diff
@@ -183,7 +183,7 @@ func (r *MulticlusterGlobalHubReconciler) Reconcile(ctx context.Context, req ctr
 	if !config.SkipDBInit(mgh) {
 		// init DB and transport here
 		if err = r.reconcileDatabase(ctx, mgh, types.NamespacedName{
-			Name:      mgh.Spec.Storage.Name,
+			Name:      mgh.Spec.DataLayer.LargeScale.Postgres.Name,
```
morvencao, line 186: nit: check if the DataLayer is nil or not.

morvencao (commented):
/lgtm

operator/pkg/controllers/hubofhubs/multiclusterglobalhub_controller.go
clyang82: will check it when support `dataLayer` is nil in the future.

PR #212, opened 2022-09-15
[ Enhance ] wait kafka and postgres secrets to be ready in e2e test

yanmxa:

### DIFF
--- a/test/setup/e2e_setup.sh
+++ b/test/setup/e2e_setup.sh
@@ -64,10 +64,10 @@ hover $! "  Enable Policy $CTX_HUB - $CTX_MANAGED"
 kubectl config use-context $CTX_HUB >> $LOG
 
 # wait kafka to be ready
-waitSecretToBeReady ${TRANSPORT_SECRET_NAME:-"transport-secret"} "open-cluster-management"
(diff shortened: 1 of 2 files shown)


test/setup/e2e_setup.sh
```diff
@@ -63,13 +63,11 @@ hover $! "  Enable Policy $CTX_HUB - $CTX_MANAGED"
 
 kubectl config use-context $CTX_HUB >> $LOG
 
-# install kafka
-bash ${CURRENT_DIR}/hoh/kafka_setup.sh $KUBECONFIG 2>&1 >> $LOG &
-hover $! "4 Install kafka cluster" 
+# wait kafka to be ready
+waitSecretToBeReady ${TRANSPORT_SECRET_NAME:-"transport-secret"} "open-cluster-management"
```
clyang82, line 67: It is working in prow environment. But it may not work in your local test environment. I would suggest to add checking the corresponding CR status or just simplify to check the pods are running.

yanmxa: Okay, I have added process to check the Postgres and Kafka cluster CRs ready.

PR #216, opened 2022-09-20
update api base path.

morvencao:


operator/pkg/controllers/hubofhubs/manifests/manager/ingress.yaml
```diff
@@ -18,5 +18,5 @@ spec:
             name: multicluster-global-hub-manager
             port:
               number: 8080
-        path: /multicloud/hub-of-hubs-nonk8s-api
+        path: /multicloud/global-hub-api
```
clyang82, line 21: We need to expose the global-hub-api via route. the management-ingress will be removed in 2.7 - https://issues.redhat.com/browse/ACM-1476

PR #222, opened 2022-09-22
[WIP]multicluster global hub API

morvencao:
Signed-off-by: morvencao [lcao@redhat.com](mailto:lcao@redhat.com)


manager/pkg/nonk8sapi/non_k8s_api.go
```diff
@@ -99,6 +90,29 @@ func AddNonK8sApiServer(mgr ctrl.Manager, database db.DB, nonK8sAPIServerConfig
...
+		clusterAPICABundle, _, err := readCertificates(nonK8sAPIServerConfig)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read certificates: %w", err)
+		}
+		router.Use(authentication.Authentication(nonK8sAPIServerConfig.ClusterAPIURL, clusterAPICABundle))
+	}
+
+	routerGroup := router.Group(nonK8sAPIServerConfig.ServerBasePath)
+	routerGroup.GET("/managedclusters", managedclusters.ListManagedClusters(database.GetConn()))
+	routerGroup.PATCH("/managedclusters/:cluster",
```
clyang82, line 108: may use clusterID to avoid duplication.

manager/pkg/nonk8sapi/managedclusters/list.go
```diff
@@ -34,16 +34,68 @@ const (
...
+		}
+
+		fmt.Fprintf(gin.DefaultWriter, "parsed selector: %s\n", selectorInSql)
+
+		limit := ginCtx.Query("limit")
+		fmt.Fprintf(gin.DefaultWriter, "limit: %v\n", limit)
+
+		lastManagedClusterName, lastManagedClusterUID := "", ""
+
+		continueToken := ginCtx.Query("continue")
```
clyang82, line 66: continue or offset?

operator/pkg/controllers/hubofhubs/manifests/manager/ingress.yaml
```diff
@@ -18,5 +18,5 @@ spec:
             name: multicluster-global-hub-manager
             port:
               number: 8080
-        path: /multicloud/hub-of-hubs-nonk8s-api
```
clyang82, line 21: Do you want to have a separated PR to remove the ingress and replace with a router?

manager/pkg/nonk8sapi/managedclusters/list.go
morvencao: Sorry for the confusion, we use `continue`, it contains a based64 encoded token for the next request to get more managed clusters in next page. It following the same pattern as kubernetes API: https://kubernetes.io/docs/reference/using-api/api-concepts/#retrieving-large-results-sets-in-chunks

operator/pkg/controllers/hubofhubs/manifests/manager/ingress.yaml
morvencao: you men create a dedicate route for global hub API?

manager/pkg/nonk8sapi/non_k8s_api.go
morvencao: the patch API still using managed cluster name and hub cluster name as parameters, I need to replace them with cluster ID.

operator/pkg/controllers/hubofhubs/manifests/manager/ingress.yaml
clyang82: Yes. Because the management-ingress will be removed in 2.7 - https://issues.redhat.com/browse/ACM-1476, that means the ingress won't be handled.

manager/pkg/nonk8sapi/managedclusters/list.go
clyang82: That is fine. You may update the document to reflect `continue` instead of `offset`.

clyang82:
@morvencao the e2e error may due to you have changed the returned structure of managedcluster. the previous structure is `[]v1.ManagedCluster`. the current structure is `[]v1.ManagedClusterList`.

PR #228, opened 2022-09-28
Fixed palcementbinding is not found error when deleting the resources from global hub

yanmxa:
Reference: https://github.com/stolostron/backlog/issues/25859


morvencao (commented):
/lgtm

PR #231, opened 2022-10-09
exclude yaml file to code coverage

clyang82:
`.yaml` is for test purpose. we should not analysis it in code coverage.


morvencao (commented):
/lgtm

PR #236, opened 2022-10-12
remove status syncers.

morvencao:
ref: https://github.com/stolostron/backlog/issues/26229


clyang82 (approved):
You may have another PR to support list appsubs so that the user can get the subscriptionID just like what we have done for the policies.

/LGTM

PR #238, opened 2022-10-13
Fixed the golangci-lint error

yanmxa:
Error message:
```bash
$ make build-manager-image
...
golangci-lint run ./cmd/... ./pkg/...
pkg/nonk8sapi/non_k8s_api_test.go:33:32: func `(*TestResponseRecorder).closeClient` is unused (unused)
func (r *TestResponseRecorder) closeClient() {
                               ^
pkg/nonk8sapi/non_k8s_api_test.go:863:45: S1025: the argument is already a string, there's no need to use fmt.Sprintf (gosimple)
                Expect(w1.Body.String()).Should(MatchJSON(fmt.Sprintf("%s", expectedPolicyStatus1)))
```

### DIFF
--- a/manager/pkg/nonk8sapi/non_k8s_api_test.go
+++ b/manager/pkg/nonk8sapi/non_k8s_api_test.go
@@ -45,6 +45,7 @@ var _ = Describe("Nonk8s API Server", Ordered, func() {
 	var postgresSQL *postgresql.PostgreSQL
 	var router *gin.Engine
 	var plc1ID string
+	var w4 *TestResponseRecorder
 
 	BeforeAll(func() {
 		var err error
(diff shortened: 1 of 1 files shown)


manager/pkg/nonk8sapi/non_k8s_api_test.go
```diff
@@ -45,6 +45,7 @@ var _ = Describe("Nonk8s API Server", Ordered, func() {
 	var postgresSQL *postgresql.PostgreSQL
 	var router *gin.Engine
 	var plc1ID string
+	var w4 *TestResponseRecorder
```
morvencao, line 48: why need a 'global' variable? Let me check it.

morvencao (commented):
/lgtm

PR #247, opened 2022-10-17
check managed clusters available for e2e tests

yanmxa:

### DIFF
--- a/test/setup/common.sh
+++ b/test/setup/common.sh
@@ -432,3 +432,21 @@ function waitPostgresToBeReady() {
   waitSecretToBeReady ${STORAGE_SECRET_NAME:-"storage-secret"} "open-cluster-management"
   echo "Postgres secret is ready"
 }
+
+function checkManagedCluster() {
+  context=$1
+  cluster=$2
+  available=$(kubectl get mcl --context "$context" | grep "$cluster" |awk '{print $5}')
+  seconds=100
+  while [[ "$available" != "True" && $seconds -gt 0 ]]; do
+    echo "retry to import managed cluster $cluster to $context"
+    initManaged $context $cluster
+    available=$(kubectl get mcl --context "$context" | grep "$cluster" |awk '{print $5}')
+    sleep 5
+    seconds=$((seconds - 5))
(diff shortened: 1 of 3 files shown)


clyang82 (commented):
I found that you just restarted the e2e tests. Do you find the root cause why we cannot have 2 mcls sometimes?

morvencao (commented):
/LGTM

yanmxa:
> 
Do you mean the issue is like this?
```
couldn't get the resource list for [operator.open-cluster-management.io/v1](http://operator.open-cluster-management.io/v1): the server could not find the requested resource
Error: no matches for kind "Klusterlet" in version "[operator.open-cluster-management.io/v1](http://operator.open-cluster-management.io/v1)"
```
As we discussed before, we have concluded that the clusteradm did not have enough time to deploy klusterlet crd. 
https://github.com/open-cluster-management-io/clusteradm/blob/257c6fd53c6c8573f2e6dafee4c826e192dcbf22/pkg/helpers/wait/wait.go#L22

clyang82:
> As we discussed before, we have concluded that the clusteradm did not have enough time to deploy klusterlet crd.

I am not if it is an environment issue or this is a common issue. You case submit a PR in clusteradm repo to see if they agree to increase the wait time. otherwise, we may add a retry here to prevent this error - https://github.com/stolostron/multicluster-global-hub/blob/0adad30e31c32e17b8def2e344e692d53c0bbfb2/test/setup/common.sh#L115

yanmxa:
> > As we discussed before, we have concluded that the clusteradm did not have enough time to deploy klusterlet crd.
> 
> I am not sure if it is an environment issue or this is a common issue. You may submit a PR in clusteradm repo to see if they agree to increase the wait time. But we can add a retry here to prevent this error -
> 
> https://github.com/stolostron/multicluster-global-hub/blob/0adad30e31c32e17b8def2e344e692d53c0bbfb2/test/setup/common.sh#L115

I'm not very clear about the issue's error message. Maybe we can hold this PR and wait for more detail about this issue from the other PRs.

PR #248, opened 2022-10-17
add oauth-proxy to nonk8s apiserver.

morvencao:
ref: https://github.com/stolostron/backlog/issues/26230

### DIFF
--- a/manager/pkg/nonk8sapi/authentication/authentication.go
+++ b/manager/pkg/nonk8sapi/authentication/authentication.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"strings"
 
 	"github.com/gin-gonic/gin"
 	userv1 "github.com/openshift/api/user/v1"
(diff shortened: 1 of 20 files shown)


operator/config/manager/manager.yaml
```diff
@@ -42,6 +42,8 @@ spec:
           value: quay.io/stolostron/multicluster-global-hub-manager:latest
         - name: OPERAND_IMAGE_MULTICLUSTER_GLOBAL_HUB_AGENT
           value: quay.io/stolostron/multicluster-global-hub-agent:latest
+        - name: OPERAND_IMAGE_OAUTH_PROXY
+          value: quay.io/stolostron/origin-oauth-proxy:4.5
```
clyang82, line 46: 4.5 is quite old. we may use 4.9 instead of 4.5. I just realized that we need to have another story to get the images from `mch-image-manifest-xxx`. I am going to log a new one.

manager/pkg/nonk8sapi/authentication/authentication.go
```diff
@@ -30,6 +31,10 @@ var errUnableToAppendCABundle = errors.New("unable to append CA Bundle")
 func Authentication(clusterAPIURL string, clusterAPICABundle []byte) gin.HandlerFunc {
```
clyang82, line 31: Do we really need it any more? Since every requests must pass the oauth-proxy authentication and we do not expose the api directly.

operator/config/manager/manager.yaml
morvencao: need to test 4.9

manager/pkg/nonk8sapi/authentication/authentication.go
morvencao: not sure whether we need it in the future for RBAC integration.

operator/config/manager/manager.yaml
morvencao: updated to 4.9

clyang82:
💯 excellent! 

/LGTM

morvencao:
ref: https://github.com/stolostron/backlog/issues/26317

PR #249, opened 2022-10-18
add controller leader election for operator and manager and agent

yanmxa:
Reference: https://github.com/stolostron/backlog/issues/26331


operator/config/manager/manager.yaml
```diff
@@ -44,6 +44,8 @@ spec:
           value: quay.io/stolostron/multicluster-global-hub-agent:latest
         - name: OPERAND_IMAGE_OAUTH_PROXY
           value: quay.io/stolostron/origin-oauth-proxy:4.9
+        - name: LEADER_ELECTION_CONFIG
```
clyang82, line 47: User cannot modify the value if it is installed from operatorhub. Is it necessary to expose it?

operator/bundle/metadata/annotations.yaml
```diff
@@ -6,7 +6,7 @@ annotations:
   operators.operatorframework.io.bundle.package.v1: multicluster-global-hub-operator
   operators.operatorframework.io.bundle.channels.v1: prerelease-2.7
   operators.operatorframework.io.bundle.channel.default.v1: prerelease-2.7
-  operators.operatorframework.io.metrics.builder: operator-sdk-v1.23.0
+  operators.operatorframework.io.metrics.builder: operator-sdk-v1.23.0+git
```
clyang82, line 9: any difference between `v1.23.0` and `v1.23.0+git`?

operator/pkg/controllers/hubofhubs/manifests/manager/deployment.yaml
```diff
@@ -29,6 +29,13 @@ spec:
             - --kafka-ssl-ca={{.KafkaCA}}
             - --process-database-url=$(DATABASE_URL)
             - --transport-bridge-database-url=$(DATABASE_URL)
+            - --cluster-api-cabundle-path=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+            - --server-certificate-path=/certs/tls.crt
+            - --server-key-path=/certs/tls.key
```
clyang82, line 34: Need to rebase your PR. those parameters are deleted.

manager/cmd/manager/main.go
```diff
@@ -158,7 +161,10 @@ func parseFlags() (*hohManagerConfig, error) {
 		"/var/run/secrets/kubernetes.io/serviceaccount/ca.crt", "The CA bundle path for cluster API.")
 	pflag.StringVar(&managerConfig.nonK8sAPIServerConfig.ServerBasePath, "server-base-path",
 		"/global-hub-api/v1", "The base path for nonK8s API server.")
-
+	pflag.IntVar(&managerConfig.electionConfig.LeaseDuration, "lease-duration", 137, "controller leader lease duration")
+	pflag.IntVar(&managerConfig.electionConfig.RenewDeadline, "renew-deadline", 107, "controller leader renew deadline")
+	pflag.IntVar(&managerConfig.electionConfig.RetryPeriod, "retry-period", 26, "controller leader retry period")
+	flag.BoolVar(&managerConfig.electionConfig.LeaderElection, "leader-elect", false, "controller leader election")
```
clyang82, line 167: why the default value is `false`?

clyang82: maybe keep it as it is. Do not expose `leader-elect`.

clyang82 (commented):
Need to support pass the leader election for agent as well.

manager/cmd/manager/main.go
yanmxa: > why the default value is `false`?

Cause the default value of leader-election in the operator is false. I set false here just to be consistent with it.

Of course, we don't have to expose it.

operator/bundle/metadata/annotations.yaml
yanmxa: I have added the `controller-leader-election-configmap `env for the operator in `operator/config/manager/manager.yaml`.

operator/config/manager/manager.yaml
yanmxa: I'm not sure whether it is a good idea to hard-code the` configmap name` to the operator code.

yanmxa: I will remove the Environment from the operator deployment.

operator/main.go
```diff
@@ -159,31 +157,30 @@ func main() {
...
-	})
+	kubeClient, err := kubernetes.NewForConfig(ctrl.GetConfigOrDie())
 	if err != nil {
-		setupLog.Error(err, "unable to start manager")
+		setupLog.Error(err, "failed to create kube client")
 		os.Exit(1)
 	}
 
-	kubeClient, err := kubernetes.NewForConfig(mgr.GetConfig())
+	electionConfig, err := getElectionConfig(kubeClient)
```
morvencao, line 166: have you tried to create the configmap that contains leader election configurations? I'm think that the client may not get that configmap since it's not added to the controller-runtime client cache.

yanmxa: I have added the configmap test in `operator/main_test.go`.

operator/main.go
```diff
@@ -205,3 +202,86 @@ func main() {
...
 	}
 }
+
+func parseFlags() *operatorConfig {
+	config := &operatorConfig{}
+	flag.StringVar(&config.MetricsAddress, "metrics-bind-address", ":8080",
+		"The address the metric endpoint binds to.")
+	flag.StringVar(&config.ProbeAddress, "health-probe-bind-address", ":8081",
+		"The address the probe endpoint binds to.")
+	flag.BoolVar(&config.LeaderElection, "leader-elect", false,
```
clyang82, line 212: suggest to use `leader-election`

operator/main.go
yanmxa: https://github.com/stolostron/multicluster-global-hub/blob/26a323a0e47890b202094d4fd8f22a75c6fde8c0/operator/main_test.go#L72

pkg/constants/constants.go
```diff
@@ -2,9 +2,10 @@ package constants
 
 const (
 	// HohSystemNamespace - Hub of Hubs dedicated namespace.
-	HohSystemNamespace  = "open-cluster-management-global-hub-system"
-	HoHConfigName       = "multicluster-global-hub-config"
-	MultiClusterHubKind = "MultiClusterHub"
+	HohSystemNamespace             = "open-cluster-management-global-hub-system"
+	HoHConfigName                  = "multicluster-global-hub-config"
+	MultiClusterHubKind            = "MultiClusterHub"
+	ControllerLeaderElectionConfig = "controller-leader-election-configmap"
```
clyang82, line 8: Add some comments for `ControllerLeaderElectionConfig`. Maybe 
```
ControllerLeaderElectionConfig allows customizing LeaseDuration /  RenewDeadline/ RetryPeriod for operator / manager / agent via configmap
```

PR #253, opened 2022-10-19
Disable subscription status and ignore hive-clusterimagesets-subscrip…

clyang82:
Disable subscription status and ignore hive-clusterimagesets-subscription-fast-0
ref: https://github.com/stolostron/backlog/issues/26321


morvencao (commented):
/LGTM

PR #260, opened 2022-10-24
remove finalizer from agent resources when deleting MGH CR

yanmxa:
Reference: https://github.com/stolostron/backlog/issues/25021


agent/pkg/jobs/purne_finalizer.go
```diff
@@ -0,0 +1,180 @@
...
+	channels := &chnv1.ChannelList{}
+	if err := p.client.List(p.ctx, channels, &client.ListOptions{}); err != nil {
+		return err
+	}
+	for idx := range channels.Items {
+		if err := p.pruneFinalizer(&channels.Items[idx]); err != nil {
+			return err
+		}
+	}
+	p.log.Info("multicluster-global-hub manager application are cleaned up")
```
morvencao, line 163: `multicluster-global-hub manager application are cleaned up` the log is misleading,  what do you think?

operator/pkg/controllers/addon/manifests/templates/hostedagent/multicluster-global-hub-hosting-agent-pre-delete-job.yaml
```diff
@@ -0,0 +1,37 @@
...
+      restartPolicy: Never
+      containers:
+      - name: multicluster-global-hub-agent
+        image: {{ .HoHAgentImage }}
+        imagePullPolicy: Always
+        args:
+          - --zap-devel=true
+          - --pod-namespace=$(POD_NAMESPACE)
+          - --leaf-hub-name={{ .LeafHubID }}
+          - --terminating=true
```
morvencao, line 29: don't  you need `kubeconfig` that access the managed cluster?

agent/pkg/jobs/purne_finalizer.go
```diff
@@ -0,0 +1,180 @@
...
+		return 1
+	}
+	if err := p.prunePolicy(); err != nil {
+		p.log.Error(err, "prune policy resources error")
+		return 1
+	}
+	return 0
+}
+
+func (p *PruneJob) pruneFinalizer(object client.Object) error {
```
clyang82, line 58: the code is duplicated. you may move it to top-level pkg.

agent/pkg/helper/config.go
```diff
@@ -102,6 +103,7 @@ func NewConfigManager() (*ConfigManager, error) {
 	pflag.IntVar(&configManager.ElectionConfig.LeaseDuration, "lease-duration", 137, "leader election lease duration")
 	pflag.IntVar(&configManager.ElectionConfig.RenewDeadline, "renew-deadline", 107, "leader election renew deadline")
 	pflag.IntVar(&configManager.ElectionConfig.RetryPeriod, "retry-period", 26, "leader election retry period")
+	pflag.BoolVar(&configManager.Terminating, "terminating", false, "PreStop addon hook")
```
clyang82, line 106: what do you mean for `terminating`? I cannot understand it thoroughly.

yanmxa: The meaning of `terminating` here is that the global-hub-agent `addon` is going to be deleted, pls do some graceful deletion. Of course, you can give a better word to describe this stage of `graceful deletion`.

agent/pkg/jobs/purne_finalizer.go
yanmxa: Thanks for your advice!

agent/pkg/helper/config.go
clyang82: I did not aware that we use the agent image to do cleanup. I am fine to use `terminating`. may just update the description with `Terminating=true is to trigger the PreStop hook to do cleanup. For example: removing finalizer`

yanmxa: Thx @zhiweiyin318!  He gave me the idea to use the agent image so that we don't need to maintain another cleanup job image for the prow.

Thx @clyang82! Good hint on the terminating flag. It seems I need to pay more attention to the program comments, making code easier to understand.

agent/pkg/jobs/purne_finalizer.go
clyang82: sorry for confusion. I am not saying move the job to the top-level pkg. the job belongs to agent. I mean in operator we have the similar logic to clean the finalizer. we may share it.
https://github.com/stolostron/multicluster-global-hub/blob/c235b4406b6c0de11c911a9fbb73d6616fc0ba33/operator/pkg/controllers/hubofhubs/finalizer.go#L50

clyang82:
💯 

/LGTM

PR #261, opened 2022-10-25
add leader election config for agent addon

yanmxa:
Reference: https://github.com/stolostron/backlog/issues/26331


zhiweiyin318:
look good to me

PR #265, opened 2022-10-28
fix hub clusterClaim is not created at the beginning

zhiweiyin318:

### DIFF
--- a/agent/pkg/controllers/hubclusterclaim_controller.go
+++ b/agent/pkg/controllers/hubclusterclaim_controller.go
@@ -1,0 +1,53 @@
+// Copyright (c) 2022 Red Hat, Inc.
+// Copyright Contributors to the Open Cluster Management project
+
+package controllers
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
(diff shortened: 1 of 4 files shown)


agent/pkg/controllers/hubclusterclaim_controller.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func (c *hubClusterClaimController) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
+	reqLogger := c.log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	reqLogger.V(2).Info("hub clusterClaim controller", request.NamespacedName)
+
+	var err error
+	mch, err := getMCH(ctx, c.client, request.NamespacedName)
+	if err != nil {
+		reqLogger.Error(err, "failed to get MCH instance")
```
clyang82, line 31: You may ignore the mch is not found error.

zhiweiyin318: check it in getMCH.

clyang82: Got it. Thanks

PR #269, opened 2022-11-01
remove deprecated label and add check clustermanager

zhiweiyin318:

### DIFF
--- a/test/setup/e2e_setup.sh
+++ b/test/setup/e2e_setup.sh
@@ -73,7 +73,6 @@ waitPostgresToBeReady
 # deploy hoh
 # need the following labels to enable deploying agent in leaf hub cluster
 kubectl label managedcluster kind-$LEAF_HUB_NAME vendor=OpenShift --overwrite 2>&1 >> $LOG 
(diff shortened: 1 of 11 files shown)


test/setup/e2e_setup.sh
```diff
@@ -73,7 +73,6 @@ waitPostgresToBeReady
 # deploy hoh
 # need the following labels to enable deploying agent in leaf hub cluster
 kubectl label managedcluster kind-$LEAF_HUB_NAME vendor=OpenShift --overwrite 2>&1 >> $LOG 
-kubectl label managedcluster kind-$LEAF_HUB_NAME global-hub.open-cluster-management.io/regional-hub-type=NoHubInstall --overwrite 2>&1 >> $LOG
```
clyang82, line 76: Can we detect OCM installed or not? I ask this question because in our e2e case, we install the OCM in the regional hub cluster.

clyang82: I am curious how does the e2e pass?

zhiweiyin318: no, we did not detect OCM, it seems like we should detect OCM too

Issue #258, opened 2022-10-21
Remove the global hub finalizer from regional hub resources when deleting the MGH CR


yanmxa:
In such a scenario, when we delete the MGH CR, we not only need to delete the agent created by the Operator on the regional hub, but we will also need to remove the global hub finalizer of other resources created by the agent, such as policy and placement.

yanmxa:
Fixed: https://github.com/stolostron/multicluster-global-hub/pull/202

yanmxa:
Global hub agent will add the finalizer back after the pre-delete addon job removes the finalizer from resources.
```bash
$ oc logs -f -l name=multicluster-global-hub-agent -n open-cluster-management-agent-addon --context regional-hub1-8xnqc
...
(exec the pre delete addon hook)
{"level":"info","ts":1667362616.2561963,"logger":"multicluster-global-hub-lease-updater","msg":"lease updater is reconciling","namespace":"open-cluster-management-agent-addon","name":"multicluster-global-hub-controller"}
{"level":"info","ts":1667362616.2622254,"logger":"multicluster-global-hub-lease-updater","msg":"lease is created or updated","namespace":"open-cluster-management-agent-addon","name":"multicluster-global-hub-controller"}
{"level":"info","ts":1667362655.730055,"logger":"placement-rules-sync","msg":"adding finalizer","Request.Namespace":"default","Request.Name":"placement-policy-limitrange"}
{"level":"info","ts":1667362655.7336943,"logger":"placement-rules-sync","msg":"Reconciliation complete.","Request.Namespace":"default","Request.Name":"placement-policy-limitrange"}
{"level":"info","ts":1667362655.7339332,"logger":"placement-rules-sync","msg":"Reconciliation complete.","Request.Namespace":"default","Request.Name":"placement-policy-limitrange"}
{"level":"info","ts":1667362655.7461772,"logger":"policies-status-sync","msg":"adding finalizer","Request.Namespace":"default","Request.Name":"policy-limitrange"}
{"level":"info","ts":1667362655.7501466,"logger":"policies-status-sync","msg":"Reconciliation complete.","Request.Namespace":"default","Request.Name":"policy-limitrange"}
{"level":"info","ts":1667362655.7502308,"logger":"policies-status-sync","msg":"Reconciliation complete.","Request.Namespace":"default","Request.Name":"policy-limitrange"}
[...]

PR #273, opened 2022-11-02
failed to remove the finalizer from resource when delete the mgh cr

yanmxa:
Fixed: https://github.com/stolostron/multicluster-global-hub/issues/258, https://github.com/stolostron/backlog/issues/26691

The process is updated to:
1. The addon pre-delete job adds a removing-finalizer label(with ttl=60s) to the resource and remove the finalizer from it
2. Global hub agent detected the removing-finalizer label in the resource and stop adding the finalizer back to it.
3. Then the finalizer is removed from resources on the regional hub.


morvencao (commented):
overall Looks good to me.
nit: the label name may be better called "GlobalHubFinalizerRemovingDeadline"?

PR #274, opened 2022-11-03
resort constants.

morvencao:
resolve: https://github.com/stolostron/backlog/issues/26697

### DIFF
--- a/agent/cmd/agent/main.go
+++ b/agent/cmd/agent/main.go
@@ -29,10 +29,10 @@ import (
 )
 
 const (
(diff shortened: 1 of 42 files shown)


clyang82:
what is the standard you are following for the constant definition? could you share it so that we can follow.

morvencao:
@clyang82 Add a new section about constants in our development.md, see: https://github.com/stolostron/multicluster-global-hub/pull/274/files#diff-a8a79d0841d496f16553c3ab96168dae29bb5ec037a9c2b45ae3f49c5a66840b

PR #275, opened 2022-11-03
replace inline kafka ca.

morvencao:
ref: https://github.com/stolostron/backlog/issues/26300

### DIFF
--- a/operator/config/samples/transport/deploy_kafka.sh
+++ b/operator/config/samples/transport/deploy_kafka.sh
@@ -66,7 +66,7 @@ bootstrapServers=$(kubectl get kafka kafka-brokers-cluster -n kafka -o jsonpath=
 kubectl get kafka kafka-brokers-cluster -n kafka -o jsonpath='{.status.listeners[1].certificates[0]}' > $currentDir/kafka-cert.pem
 kubectl create secret generic ${transportSecret} -n $TARGET_NAMESPACE \
     --from-literal=bootstrap_server=$bootstrapServers \
-    --from-file=CA=$currentDir/kafka-cert.pem
+    --from-file=ca.crt=$currentDir/kafka-cert.pem
 
 rm $currentDir/kafka-cert.pem
 echo "transport secret is ready in $TARGET_NAMESPACE namespace!"
--- a/operator/pkg/controllers/addon/manifests/templates/hostedagent/multicluster-global-hub-hosting-agent-deployment.yaml
+++ b/operator/pkg/controllers/addon/manifests/templates/hostedagent/multicluster-global-hub-hosting-agent-deployment.yaml
@@ -28,7 +28,7 @@ spec:
             - --enforce-hoh-rbac=false
             - --transport-type=kafka
             - --kafka-bootstrap-server={{ .KafkaBootstrapServer }}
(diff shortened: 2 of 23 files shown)


operator/config/samples/transport/deploy_kafka.sh
```diff
@@ -66,7 +66,7 @@ bootstrapServers=$(kubectl get kafka kafka-brokers-cluster -n kafka -o jsonpath=
 kubectl get kafka kafka-brokers-cluster -n kafka -o jsonpath='{.status.listeners[1].certificates[0]}' > $currentDir/kafka-cert.pem
 kubectl create secret generic ${transportSecret} -n $TARGET_NAMESPACE \
     --from-literal=bootstrap_server=$bootstrapServers \
-    --from-file=CA=$currentDir/kafka-cert.pem
+    --from-file=ca.crt=$currentDir/kafka-cert.pem
```
yanmxa, line 69: Since we update the transport secret key `CA` to  `ca.crt`. This line should also be changed to `ca.cert`
https://github.com/stolostron/multicluster-global-hub/blob/f1a14473f00d849364bf77b00ab43cf3747c7fea/operator/pkg/utils/utils.go#L100

operator/pkg/controllers/addon/manifests/templates/hostedagent/multicluster-global-hub-hosting-agent-deployment.yaml
```diff
@@ -28,7 +28,7 @@ spec:
             - --enforce-hoh-rbac=false
             - --transport-type=kafka
             - --kafka-bootstrap-server={{ .KafkaBootstrapServer }}
-            - --kafka-ssl-ca={{ .KafkaCA }}
+            - --kafka-ssl-ca=path=/kafka-ca/ca.crt
```
yanmxa, line 31: How about mounting the Kafka secret `ca.cert` to the operator deployment? I'm not sure if this is a good idea.

operator/config/samples/transport/deploy_kafka.sh
morvencao: already updated that, see: https://github.com/stolostron/multicluster-global-hub/pull/275/files#diff-ff8fd2296cfd658ed6708d8e58d8e3ac804532b5ae0917d43d44492606b27abb

operator/pkg/controllers/addon/manifests/templates/hostedagent/multicluster-global-hub-hosting-agent-deployment.yaml
morvencao: Since we finally need to let the manager and agent to mount the kafka, the operator just checks what is provided in MGH CR(transport-secret) and pass it to manager and agent.

Image a case, where a user install operator, but no kafka secret and MGH CR are created yet, the operator will not be started.

yanmxa: Yes! It's important to keep the transport information in the MGH CR, not operator.

clyang82:
It looks good for me. Please @yanmxa take another look.

PR #276, opened 2022-11-03
add OCM prefix for clusterrole and clusterrolebinding

yanmxa:
Referrence: https://github.com/stolostron/backlog/issues/26271

### DIFF
--- a/operator/config/rbac/auth_proxy_client_clusterrole.yaml
+++ b/operator/config/rbac/auth_proxy_client_clusterrole.yaml
@@ -1,0 +1,9 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: metrics-reader
+rules:
+- nonResourceURLs:
+  - "/metrics"
+  verbs:
+  - get
--- a/operator/config/rbac/auth_proxy_client_clusterrole.yaml
+++ b/operator/config/rbac/auth_proxy_client_clusterrole.yaml
@@ -1,7 +1,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: metrics-reader
+  name: open-cluster-management:metrics-reader
 rules:
 - nonResourceURLs:
   - "/metrics"
--- a/operator/config/rbac/auth_proxy_client_clusterrole.yaml
+++ b/operator/config/rbac/auth_proxy_client_clusterrole.yaml
@@ -1,9 +1,0 @@
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
-metadata:
-  name: open-cluster-management:metrics-reader
-rules:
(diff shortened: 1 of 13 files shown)


operator/config/rbac/auth_proxy_client_clusterrole.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: metrics-reader
+  name: open-cluster-management:metrics-reader
```
clyang82, line 4: we do not use this role. @morvencao right?

morvencao: correct, we will never use this role, so no need to update it. @yanmxa

morvencao (changes requested):
@yanmxa I think we don't add prefix to the role/clusterrole/rolebing/clusterrolebinding to the operator, because they will be installed with OLM, the csv will help to generate the names. We should focus on role/clusterrole/rolebing/clusterrolebinding for manager and agent? right? @clyang82

operator/config/rbac/auth_proxy_client_clusterrole.yaml
yanmxa: Yes, I haven't seen this `clusterrole` is bound to any object. And I will remove it

morvencao (commented):
Also I think you missed role and rolebinding for agent in hosted mode.

yanmxa:
> Also I think you missed role and rolebinding for agent in hosted mode.

Maybe we just need to add the `open-cluster-management` to the `clusterrole` and `clusterrolebinding`.  Cause role and rolbinding is namespaced. We don't need a prefix to indicate where it's from. 

Am I right? @clyang82 @morvencao

yanmxa:
> @yanmxa I think we don't add prefix to the role/clusterrole/rolebing/clusterrolebinding to the operator, because they will be installed with OLM, the csv will help to generate the names. We should focus on role/clusterrole/rolebing/clusterrolebinding for manager and agent? right? @clyang82

Agree!

morvencao:
OK @yanmxa make sense to me! thanks.

operator/config/rbac/auth_proxy_client_clusterrole.yaml
```diff
@@ -1,9 +0,0 @@
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
-metadata:
-  name: metrics-reader
-rules:
-- nonResourceURLs:
-  - "/metrics"
-  verbs:
-  - get
```
morvencao, line 9: Let's keep it for future usage.

yanmxa: Okay. I rollback the role.

morvencao (approved):
/lgtm

PR #284, opened 2022-11-07
install the agent in the global-hub-system ns

zhiweiyin318:

### DIFF
--- a/agent/pkg/controllers/controllers_integration_test.go
+++ b/agent/pkg/controllers/controllers_integration_test.go
@@ -80,7 +80,7 @@ var _ = Describe("controller", Ordered, func() {
 			}, clusterClaim)
 		}, 1*time.Second, 100*time.Millisecond).ShouldNot(HaveOccurred())
 		Expect(clusterClaim.Spec.Value).Should(Equal(
-			constants.HubInstalledWithoutSelfManagement))
+			constants.HubInstalledByUser))
 	})
 
 	It("clusterClaim testing clusterManager and mch are not installed", func() {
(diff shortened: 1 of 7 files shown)


pkg/constants/constants.go
```diff
@@ -55,10 +55,9 @@ const (
 	HubClusterClaimName = "hub.open-cluster-management.io"
 
 	// the value of the HubClusterClaimName ClusterClaim
-	HubNotInstalled                   = "NotInstalled"
-	HubInstalledWithSelfManagement    = "InstalledWithSelfManagement"
-	HubInstalledWithoutSelfManagement = "InstalledWithoutSelfManagement"
-	HubInstalledByHoH                 = "InstalledByHoH"
+	HubNotInstalled    = "NotInstalled"
+	HubInstalledByUser = "InstalledByUser"
+	HubInstalledByHoH  = "InstalledByHoH"
```
clyang82, line 60: suggest to use `InstalledByGlobalHub`

agent/pkg/controllers/controllers_integration_test.go
```diff
@@ -80,7 +80,7 @@ var _ = Describe("controller", Ordered, func() {
 			}, clusterClaim)
 		}, 1*time.Second, 100*time.Millisecond).ShouldNot(HaveOccurred())
 		Expect(clusterClaim.Spec.Value).Should(Equal(
-			constants.HubInstalledWithoutSelfManagement))
+			constants.HubInstalledByUser))
```
clyang82, line 83: that means we do not care about if the hub cluster is installed with self management or not. correct?

pkg/constants/constants.go
zhiweiyin318: done

agent/pkg/controllers/controllers_integration_test.go
zhiweiyin318: yes. we do not need to detect the local-cluster.

PR #285, opened 2022-11-09
fix incarnation in hosted mode.

morvencao:
ref: https://issues.redhat.com/browse/ACM-2023

### DIFF
--- a/agent/cmd/agent/main.go
+++ b/agent/cmd/agent/main.go
@@ -170,7 +170,7 @@ func createManager(restConfig *rest.Config, environmentManager *helper.ConfigMan
 	}
 
 	// incarnation version
-	incarnation, err := incarnation.GetIncarnation(mgr, environmentManager.PodNameSpace)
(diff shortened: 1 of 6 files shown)


yanmxa:
We also need to parse the flag parameter before using the `ctrl.GetConfigOrDie()`. 
https://github.com/stolostron/multicluster-global-hub/blob/a3bb618aaf32f7bfe1cf9504ad0996fca49c33a5/agent/cmd/agent/main.go#L39

Otherwise, It might not get the [kubeconfig](https://github.com/stolostron/multicluster-global-hub/blob/a3bb618aaf32f7bfe1cf9504ad0996fca49c33a5/operator/pkg/controllers/addon/manifests/templates/hostedagent/multicluster-global-hub-hosting-agent-deployment.yaml#L24) flag when executing the `ctrl.GetConfigOrDie()`

morvencao:
will fix that in another PR

PR #288, opened 2022-11-16
add UT for transport to db

yanmxa:
Reference: https://github.com/stolostron/backlog/issues/26233


morvencao (commented):
/lgtm

PR #289, opened 2022-11-23
fix the addon deletion when mgh is deleted

zhiweiyin318:
issue: https://issues.redhat.com/browse/ACM-2095


morvencao (commented):
/lgtm

PR #290, opened 2022-12-05
install agent in klusterlet ns when hosted mode

zhiweiyin318:


operator/pkg/controllers/hubofhubs/manifests/manager/managementaddon.yaml
```diff
@@ -5,6 +5,8 @@ metadata:
   name: multicluster-global-hub-controller
   labels:
     name: multicluster-global-hub-manager
+  annotations:
+    skip-creation-if-exist: "true"
```
morvencao, line 9: this annotation means it can't be updated after creation., is this what you want?

zhiweiyin318: yes,  some new features of addon require  managementaddon to be updated manually.

morvencao (commented):
/lgtm

PR #291, opened 2022-12-06
fix addon work rbac

zhiweiyin318:
1. add rbac to work-agent to apply operatorgroup and subscription.
2. update addon-framework to fix annotation typo.
3. to delete hoh addon cr when deploy-mode is None.


morvencao (commented):
/lgtm

PR #295, opened 2022-12-09
use latest clusteradm in e2e test setup.

morvencao:


yanmxa:
@morvencao Why do we use the latest version of `clusteradm` instead of a pinned version?

morvencao:
add version `0.4.1` for clusteradm install script.

PR #259, opened 2022-10-24
refactor the transport

clyang82:
ref: https://github.com/stolostron/backlog/issues/26483
Signed-off-by: clyang82 <chuyang@redhat.com>

PR #305, opened 2023-01-05
fix duplicated kafka consumer ID in global hub agent.

morvencao:
fix: https://github.com/stolostron/backlog/issues/27083
which is introduced in https://github.com/stolostron/multicluster-global-hub/pull/259/

### DIFF
--- a/operator/pkg/controllers/addon/manifests/templates/agent/multicluster-global-hub-agent-deployment.yaml
+++ b/operator/pkg/controllers/addon/manifests/templates/agent/multicluster-global-hub-agent-deployment.yaml
@@ -24,6 +24,7 @@ spec:
             - '--zap-devel=true'
             - --pod-namespace=$(POD_NAMESPACE)
             - --leaf-hub-name={{ .LeafHubID }}
+            - --kakfa-consumer-id={{ .LeafHubID }}
             - --enforce-hoh-rbac=false
             - --transport-type=kafka
             - --kafka-bootstrap-server={{ .KafkaBootstrapServer }}
--- a/agent/cmd/agent/main_test.go
+++ b/agent/cmd/agent/main_test.go
@@ -116,6 +116,8 @@ func TestAgent(t *testing.T) {
 			"default",
 			"--leaf-hub-name",
 			"hub1",
+			"--kakfa-consumer-id",
+			"hub1",
 			"--lease-duration",
 			"15",
 			"--renew-deadline",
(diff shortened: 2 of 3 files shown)


operator/pkg/controllers/addon/manifests/templates/agent/multicluster-global-hub-agent-deployment.yaml
```diff
@@ -24,6 +24,7 @@ spec:
             - '--zap-devel=true'
             - --pod-namespace=$(POD_NAMESPACE)
             - --leaf-hub-name={{ .LeafHubID }}
+            - --kakfa-consumer-id={{ .LeafHubID }}
```
clyang82, line 27: what is default value without specifying it obviously?

yanmxa: Each consumer has a unique consumer.id across its consumer group. If you don't define a consumer.id, the Kafka client (Java, Python, Node.js, etc) usually chooses a random ID.
Refer: https://stackoverflow.com/questions/34550873/difference-between-groupid-and-consumerid-in-kafka-consumer

yanmxa:
I think maybe we need to make sure different regional hub have a different `group.id`. 
image

As you can see from this figure, if two consumers are in the same group, there is no guarantee that each consumer in the group will receive the message, but every group will receive the full information about the topic.

operator/pkg/controllers/addon/manifests/templates/agent/multicluster-global-hub-agent-deployment.yaml
morvencao: we use `kakfa-consumer-id` as kafka consumer ID and group ID, see: https://github.com/stolostron/multicluster-global-hub/blob/7dcb96485c9576e26d8a875134cb1a095e87b17e/pkg/transport/consumer/kafka_consumer.go#L69-L70 That means each agent can receive all the partitions of the message.

morvencao:
we use `kakfa-consumer-id` as kafka consumer ID and group ID, see: https://github.com/stolostron/multicluster-global-hub/blob/7dcb96485c9576e26d8a875134cb1a095e87b17e/pkg/transport/consumer/kafka_consumer.go#L69-L70 That means each agent can receive all the partitions of the message.
@yanmxa

clyang82:
It was introduced by https://github.com/stolostron/multicluster-global-hub/pull/259. the previous value is regional hub name. #259 made it as an immutable value.

/lgtm

PR #311, opened 2023-02-01
add timestamp for status tables

yanmxa:


yanmxa:
Consider the scale performance. The timestamp trigger seems not a good choice.

PR #313, opened 2023-02-09
update version.

morvencao:


morvencao:
Looks like e2e workflow `ocm-e2e-kind` is broken now, asked cicd for help.

PR #314, opened 2023-02-09
fix oauth-proxy permission issue.

morvencao:
See: https://github.com/openshift/oauth-proxy#limiting-access-to-users for `--openshift-sar` usage
End user or service account must have permission list in `--openshift-sar` to access backend service.

Newly created user doesn't have permission to get  or list `namespaces`:

```bash
# oc get ns
Error from server (Forbidden): namespaces is forbidden: User "acm-user1" cannot list resource "namespaces" in API group "" at the cluster scope
```

But newly created user has permission to list `projects` in openshift, but can't see any project:

```
# oc get projects
No resources found
```


clyang82:
🥇 
/lgtm

PR #318, opened 2023-02-15
Replace transport spec path with cloudevents

yanmxa:


agent/pkg/status/controller/managedclusters/clusters_status_sync.go
```diff
@@ -34,10 +34,11 @@ func AddClustersStatusController(mgr ctrl.Manager, producer producer.Producer, l
 		})
 	}
 
-	predicateFunc := func() bool { // bundle predicate
-		return hubOfHubsConfig.Data["aggregationLevel"] == "full" ||
-			hubOfHubsConfig.Data["aggregationLevel"] == "minimal"
+	predicateFunc := func() bool {
+		// return hubOfHubsConfig.Data["aggregationLevel"] == "full" ||
+		// 	hubOfHubsConfig.Data["aggregationLevel"] == "minimal"
 		// at this point send all managed clusters even if aggregation level is minimal
```
clyang82, line 40: Can we continue supporting this configuration?

pkg/transport/consumer/generic_consumer.go
```diff
@@ -0,0 +1,94 @@
...
+	var receiver interface{}
+	switch transportConfig.TransportType {
+	case string(transport.Kafka):
+		var err error
+		log.Info("transport consumer with kafka receiver")
+		receiver, err = protocol.NewKafkaReceiver(transportConfig.KafkaConfig)
+		if err != nil {
+			return nil, err
+		}
+	case string(transport.Chan):
```
clyang82, line 38: Does it for test only?

pkg/transport/producer/generic_producer.go
```diff
@@ -0,0 +1,104 @@
...
+
+type GenericProducer struct {
+	log              logr.Logger
+	client           cloudevents.Client
+	messageSizeLimit int
+}
+
+func NewGenericProducer(transportConfig *transport.TransportConfig) (*GenericProducer, error) {
+	var sender interface{}
+	messageSize := 960 * 1000
```
clyang82, line 31: Can we make it configurable?

clyang82 (changes requested):

pkg/transport/consumer/generic_consumer.go
yanmxa: Yeah, this is for unit test.

pkg/transport/producer/generic_producer.go
yanmxa: It is configurable. the` messageSize := 960 * 1000` is the default value. Actually, it reads the config value from the below line `messageSize = transportConfig.KafkaConfig.ProducerConfig.MessageSizeLimitKB * 1000`

agent/pkg/status/controller/managedclusters/clusters_status_sync.go
yanmxa: The `aggregationLevel` is only for policies, it should return `true` for the managed cluster

PR #319, opened 2023-02-16
replace transport status path with cloudevents

yanmxa:
Add a `transportFomat` to provide an option to enable `cloudevents` or not fro customer.
```yaml
apiVersion: operator.open-cluster-management.io/v1alpha2
kind: MulticlusterGlobalHub
metadata:
  annotations:
    mgh-image-overrides-cm: mgh-images-config
  name: multiclusterglobalhub
spec:
  dataLayer:
    type: largeScale
    largeScale:
      kafka:
        name: ${TRANSPORT_SECRET_NAME}
        transportFormat: message # or cloudEvents
      postgres:
        name: ${STORAGE_SECRET_NAME}
```


clyang82:
I would like to expose `transport_format` in API. the values can be `message` and `cloudevents`. the default is `cloudevents`. what do you think?

yanmxa:
> I would like to expose `transport_format` in API. the values can be `message` and `cloudevents`. the default is `cloudevents`. what do you think?

That is a grea

> mgh-cloudevents

Good idea! we can expose this option to the customer. I will update the API like this:
```yaml
apiVersion: operator.open-cluster-management.io/v1alpha2
kind: MulticlusterGlobalHub
metadata:
  annotations:
    mgh-image-overrides-cm: mgh-images-config
  name: multiclusterglobalhub
spec:
  dataLayer:
    type: largeScale
    largeScale:
      kafka:
        name: ${TRANSPORT_SECRET_NAME}
        cloudevents: "true" # or false
      postgres:
        name: ${STORAGE_SECRET_NAME}
 ```

**Review by morvencao — COMMENTED** [review]
I think @clyang82 is suggesting to change API to something like this:
```yaml
apiVersion: operator.open-cluster-management.io/v1alpha2
kind: MulticlusterGlobalHub
metadata:
  annotations:
    mgh-image-overrides-cm: mgh-images-config
  name: multiclusterglobalhub
spec:
  dataLayer:
    type: largeScale
    largeScale:
      kafka:
        name: ${TRANSPORT_SECRET_NAME}
        messageType: kafkaMessage # or cloudEvents
      postgres:
        name: ${STORAGE_SECRET_NAME}
```

If we have `messageType` field, it is easy to add a new  message type. If we use `cloudevents`(true or false) knob,  the API change will be hard in the future.

yanmxa:
> I think @clyang82 is suggesting to change API to something like this:
> 
> ```yaml
> apiVersion: operator.open-cluster-management.io/v1alpha2
> kind: MulticlusterGlobalHub
> metadata:
>   annotations:
>     mgh-image-overrides-cm: mgh-images-config
>   name: multiclusterglobalhub
> spec:
>   dataLayer:
>     type: largeScale
>     largeScale:
>       kafka:
>         name: ${TRANSPORT_SECRET_NAME}
>         messageType: kafkaMessage # or cloudEvents
>       postgres:
>         name: ${STORAGE_SECRET_NAME}
> ```
> 
> If we have `messageType` field, it is easy to add a new message type. If we use `cloudevents`(true or false) knob, the API change will be hard in the future.

@morvencao I agree with you that the `messageType(string)` field might be flexible to add another value, but for the Kafka transport, it seems more readable to add a `boolean` value to let the customer decide to use `cloudevent` or not. Because in terms of Kafka scenarios, I don't think there are other message types to consider yet. Anyway, thanks for the suggestion, let's discuss tomorrow which way is more suitable for our scenario.

manager/pkg/statussyncer/transport2db/syncer/syncers.go
```diff
@@ -5,37 +5,100 @@ import (
...
+			conflationManager.GetBundlesMetadata, ctrl.Log.WithName("kafka-consumer")),
+		)
+		kafkaConsumer.SetStatistics(stats)
+		if err := mgr.Add(kafkaConsumer); err != nil {
+			return nil, fmt.Errorf("failed to add status transport bridge: %w", err)
+		}
+		bundleRegisterable = kafkaConsumer
+	} else {
+		// create consumer
+		consumer, err := consumer.NewGenericConsumer(managerConfig.TransportConfig)
```
clyang82, line 86: is it for sync-service?

manager/pkg/statussyncer/transport2db/syncer/syncers.go
```diff
@@ -50,11 +113,55 @@ func AddTransport2DBSyncers(mgr ctrl.Manager, dbWorkerPool *workerpool.DBWorkerP
...
 
-	return nil
+	return bundleRegisterable, nil
+}
+
+// function to determine whether the transport component requires initial-dependencies between bundles to be checked
+// (on load). If the returned is false, then we may assume that dependency of the initial bundle of
+// each type is met. Otherwise, there are no guarantees and the dependencies must be checked.
+func requireInitialDependencyChecks(transportType string) bool {
+	switch transportType {
```
clyang82, line 127: what is the value for transport Type? I guess it should be `kafka` only right now.

pkg/transport/producer/generic_producer.go
```diff
@@ -26,11 +26,11 @@ type GenericProducer struct {
 	messageSizeLimit int
 }
 
-func NewGenericProducer(transportConfig *transport.TransportConfig) (*GenericProducer, error) {
+func NewGenericProducer(transportConfig *transport.TransportConfig) (transport.Producer, error) {
 	var sender interface{}
 	messageSize := 960 * 1000
 	switch transportConfig.TransportType {
-	case string(transport.Kafka):
+	case string(transport.Cloudevents), string(transport.Kafka):
```
clyang82, line 33: how do you handle the different transport format: message & cloudevents? I saw right now we only handle the cloudevents. Can you point me out for handling kafka message?

manager/pkg/statussyncer/transport2db/syncer/syncers.go
yanmxa: it is for cloudevents.

pkg/transport/producer/generic_producer.go
yanmxa: We will always use `cloudevents` on `transport spec path`. So the spec producer will always use the cloudevents client whether `transportType` is `transport.Cloudevents` or `transport.Kafka`

This is the status path kafka producer: https://github.com/yanmxa/hub-of-hubs/blob/0881caf364461484e6d45eea77dbe3874a6ae2ef/agent/pkg/status/controller/controller.go#L41-L58
And this is the status path kafka consumer: https://github.com/yanmxa/hub-of-hubs/blob/0881caf364461484e6d45eea77dbe3874a6ae2ef/manager/pkg/statussyncer/transport2db/syncer/syncers.go#L63-L83

manager/pkg/statussyncer/transport2db/syncer/syncers.go
yanmxa: For now, the value of transport Type is either `kafka` or `cloudevents-kaka`.

yanmxa:
> I think @clyang82 is suggesting to change API to something like this:
> 
> ```yaml
> apiVersion: operator.open-cluster-management.io/v1alpha2
> kind: MulticlusterGlobalHub
> metadata:
>   annotations:
>     mgh-image-overrides-cm: mgh-images-config
>   name: multiclusterglobalhub
> spec:
>   dataLayer:
>     type: largeScale
>     largeScale:
>       kafka:
>         name: ${TRANSPORT_SECRET_NAME}
>         messageType: kafkaMessage # or cloudEvents
>       postgres:
>         name: ${STORAGE_SECRET_NAME}
> ```
> 
> If we have `messageType` field, it is easy to add a new message type. If we use `cloudevents`(true or false) knob, the API change will be hard in the future.

@clyang82 @morvencao Just as we discussed. I updated the MGH API to this:
```yaml
apiVersion: operator.open-cluster-management.io/v1alpha2
kind: MulticlusterGlobalHub
metadata:
  annotations:
    mgh-image-overrides-cm: mgh-images-config
  name: multiclusterglobalhub
spec:
  dataLayer:
    type: largeScale
    largeScale:
      kafka:
        name: ${TRANSPORT_SECRET_NAME}
        format: kafkaMessage # or cloudEvents
      postgres:
        name: ${STORAGE_SECRET_NAME}
```

manager/pkg/statussyncer/transport2db/syncer/syncers.go
clyang82: maybe use messageType to align with the definition.

clyang82: the values are `message` and `cloudevents`

clyang82:
```
apiVersion: operator.open-cluster-management.io/v1alpha2
kind: MulticlusterGlobalHub
metadata:
  annotations:
    mgh-image-overrides-cm: mgh-images-config
  name: multiclusterglobalhub
spec:
  dataLayer:
    type: largeScale
    largeScale:
      kafka:
        name: ${TRANSPORT_SECRET_NAME}
        transportFormat: message # or cloudEvents
      postgres:
        name: ${STORAGE_SECRET_NAME}
```

manager/pkg/statussyncer/transport2db/syncer/syncers.go
yanmxa: Done!

yanmxa:
> ```
> apiVersion: operator.open-cluster-management.io/v1alpha2
> kind: MulticlusterGlobalHub
> metadata:
>   annotations:
>     mgh-image-overrides-cm: mgh-images-config
>   name: multiclusterglobalhub
> spec:
>   dataLayer:
>     type: largeScale
>     largeScale:
>       kafka:
>         name: ${TRANSPORT_SECRET_NAME}
>         transportFormat: message # or cloudEvents
>       postgres:
>         name: ${STORAGE_SECRET_NAME}
> ```

@clyang82 Done!

operator/config/crd/bases/operator.open-cluster-management.io_multiclusterglobalhubs.yaml
```diff
@@ -56,13 +56,17 @@ spec:
                       use postgres as data layer This is for a large scale environment.
                     properties:
                       kafka:
-                        description: LocalObjectReference contains enough information
-                          to let you locate the referenced object inside the same
-                          namespace.
+                        description: KafkaConfig is the config of kafka
```
clyang82, line 59: suggest to use `kafka defines the desired state of kafka`

operator/config/crd/bases/operator.open-cluster-management.io_multiclusterglobalhubs.yaml
```diff
@@ -56,13 +56,17 @@ spec:
...
                         properties:
                           name:
-                            description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
-                              TODO: Add other useful fields. apiVersion, kind, uid?'
+                            type: string
+                          transportFormat:
+                            default: cloudEvents
+                            description: This flag tells the controller to using the
```
clyang82, line 65: `It defines the transport format for kafka`

clyang82 (approved):
/lgtm

You may address the minor comments in another PR. Since this PR is a big PR, we can merge it firstly.

PR #322, opened 2023-02-22
blocking the addon installation when MGH CR has the pause annotation

yanmxa:


morvencao (commented):
/lgtm

PR #329, opened 2023-02-28
enable auto commit for the cloudevents

yanmxa:

If we set auto commit to false, it will consume the message from the beginning when restarting the consumer every time.

we need to make sure the consumer receives the message from the last committed offset. Otherwise, after the global hub restarts, it will consume a large number of duplicate bundles.


morvencao (commented):
currently when the consumer restarts, does the committer will remember the commit offset?

yanmxa:
> currently when the consumer restarts, does the committer will remember the commit offset?

Yes, It will consume messages from the last committed offset.

PR #334, opened 2023-03-03
correct the local policy e2e test

yanmxa:

### DIFF
--- a/test/pkg/e2e/hoh-local-policy_test.go
+++ b/test/pkg/e2e/hoh-local-policy_test.go
@@ -2,9 +2,7 @@ package tests
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
-	"strings"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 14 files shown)


test/pkg/e2e/hoh-local-policy_test.go
```diff
@@ -31,136 +32,181 @@ const (
...
-					utils.NewPolicyGVR()).Namespace(LOCAL_POLICY_NAMESPACE).Get(
-					context.TODO(), LOCAL_POLICY_NAME, metav1.GetOptions{})
-				Expect(err).ShouldNot(HaveOccurred())
-				unstructedContent := unstrcutedPolicy.UnstructuredContent()
-				var policy policiesv1.Policy
-				err = runtime.DefaultUnstructuredConverter.FromUnstructured(unstructedContent, &policy)
-				Expect(err).ShouldNot(HaveOccurred())
-				if len(policy.Status.Status) <= 0 {
-					return fmt.Errorf("inform local policy status is not ready")
+				policyStr, err := clients.Kubectl(clients.HubClusterName(), "exec", "-i", postgresMaster, "-c", "database",
```
morvencao, line 166: why not use postgresql client to access DB?

yanmxa: Currently, the host of `postgreURI` is made of `clusterIP`, which means we can't access it from outside the cluster. If we use the postgresql client, we have to create another `NodePort` service to expose it.

yanmxa: @morvencao I have started a `nodeport` of the postgres cluster to enable the PostgreSQL client.

morvencao (commented):
/lgtm

PR #335, opened 2023-03-03
fixed the security hotspots from the sunar postsubmit

yanmxa:
Risk message: Scratch images run as root by default. Make sure it is safe here.


morvencao (commented):
/lgtm

PR #336, opened 2023-03-06
fixed the security issue of the sonar 'defer'

yanmxa:
Error message: `Deferring unsafe method "Close" on type "io.ReadCloser"`


morvencao (commented):
in this case, we just print the error message?

PR #342, opened 2023-03-09
add postgres connection(`ca.crt`) for grafana datasource

yanmxa:

### Set the grafana datasource by console:
**Scenario 1**: If we set the`sslmode = "requre" (or prefer)`, then don't need any Postgres certs for the connection.
Scenario 2: If we set the `sslmode="disable"`, then it will throw "no encryption" error.
**Scenario 3**: If we set the `sslmode="verify-ca"(or verify-full)`, then we need to provide `ca.crt`.
### Provision the grafana datasource by global hub operator:
**Scenario 4**: If we set the `sslmode="require"(or prefer)`, then don't need any Postgres certs for the connection.
Scenario 5: If we set the `sslmode="disable"`, then it will throw "no encryption" instead of being connected.
Scenario 6: If we set the `sslmode="verify-ca"(or verify-full)`, the storage secret need to provide `ca.crt`.

Note: the grafana data source connection enables the SSL by default.

refs: https://jdbc.postgresql.org/documentation/ssl/


operator/config/samples/storage/deploy_postgres.sh
```diff
@@ -26,12 +26,20 @@ waitAppear "kubectl get secret hoh-pguser-postgres -n hoh-postgres --ignore-not-
...
+
+kubectl get secret $certSecret -n $pgnamespace -o jsonpath='{.data.ca\.crt}' |base64 -d > $currentDir/ca.crt
+kubectl get secret $certSecret -n $pgnamespace -o jsonpath='{.data.tls\.crt}' |base64 -d > $currentDir/tls.crt
+kubectl get secret $certSecret -n $pgnamespace -o jsonpath='{.data.tls\.key}' |base64 -d > $currentDir/tls.key
 
 kubectl create secret generic $storageSecret -n $targetNamespace \
-    --from-literal=database_uri="postgres://${databaseUser}:${databasePassword}@${databaseHost}:${databasePort}/hoh"
+    --from-literal=database_uri="postgres://${databaseUser}:${databasePwd}@${databaseHost}:${databasePort}/hoh" \
+    --from-file=ca.crt=$currentDir/ca.crt \
+    --from-file=tls.crt=$currentDir/tls.crt \
```
clyang82, line 43: Why we need both basic auth and tls auth? I think you only need `ca.crt` here.

morvencao: Is it multls?

yanmxa: I verified the grafana datasource connection, and found it enable the SSL by default. That means we even don't need to provide the `ca.crt`. We just need to make sure the postgres server enabled the SSL and the storage secret only keeps the `database_uri` without other `crts`.

yanmxa:
Sample postgres datasources for grafana
```json
{
    "id": 11,
    "uid": "S2e2qo-4k",
    "orgId": 1,
    "name": "hello-verify-ca",
    "type": "postgres",
    "typeName": "PostgreSQL",
    "typeLogoUrl": "public/app/plugins/datasource/postgres/img/postgresql_logo.svg",
    "access": "proxy",
    "url": "myan.hosts.prod.psi.pek2.redhat.com:32000",
    "user": "hippo",
    "database": "hippo",
    "basicAuth": false,
    "isDefault": false,
    "jsonData": {
      "postgresVersion": 903,
      "sslmode": "verify-ca",
      "tlsAuth": true,
      "tlsAuthWithCACert": true,
      "tlsConfigurationMethod": "file-content",
      "tlsSkipVerify": false
    },
    "readOnly": false
  }
```

morvencao (commented):
/lgtm

PR #344, opened 2023-03-14
verify the local policy resources without the global finalizer

yanmxa:

There are three main roles for the cleanup finalizer:

1. the global hub manager uses it to [clean up the data in the database when deleting this resource on the global hub](https://github.com/stolostron/multicluster-global-hub/blob/bc66f7a075a649aedc30fe7edca7ec858307f21d/manager/pkg/specsyncer/spec2db/controller/generic_spec_to_db_reconciler.go#L83-L84).
2. the global hub operator uses it to clean up the resources before removing the MGH CR.
3. the global hub agent needs the finalizer to remove the [deleting obj from the bundle](https://github.com/stolostron/multicluster-global-hub/blob/bc66f7a075a649aedc30fe7edca7ec858307f21d/agent/pkg/status/controller/generic/generic_status_sync_controller.go#L135-L141). So the deleted obj can be propagated to the database.

### DIFF
--- a/agent/pkg/status/controller/generic/generic_status_sync_controller.go
+++ b/agent/pkg/status/controller/generic/generic_status_sync_controller.go
@@ -16,6 +16,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
+	"github.com/stolostron/multicluster-global-hub/agent/pkg/helper"
 	"github.com/stolostron/multicluster-global-hub/agent/pkg/status/bundle"
 	"github.com/stolostron/multicluster-global-hub/agent/pkg/status/controller/syncintervals"
 	"github.com/stolostron/multicluster-global-hub/pkg/constants"
@@ -226,6 +227,10 @@ func (c *genericStatusSyncController) addFinalizer(ctx context.Context, object b
 	}
 
 	if controllerutil.ContainsFinalizer(object, c.finalizerName) {
+		return nil
+	}
+	// if local policy resources, then skip the adding finalizer action
+	if !helper.HasAnnotation(object, constants.OriginOwnerReferenceAnnotation) {
 		return nil
 	}
 
--- a/agent/pkg/status/controller/generic/generic_status_sync_controller.go
+++ b/agent/pkg/status/controller/generic/generic_status_sync_controller.go
@@ -16,7 +16,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
-	"github.com/stolostron/multicluster-global-hub/agent/pkg/helper"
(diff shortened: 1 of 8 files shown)


agent/pkg/status/controller/generic/generic_status_sync_controller.go
```diff
@@ -228,6 +229,10 @@ func (c *genericStatusSyncController) addFinalizer(ctx context.Context, object b
 	if controllerutil.ContainsFinalizer(object, c.finalizerName) {
 		return nil
 	}
+	// if local policy resources, then skip the adding finalizer action
```
morvencao, line 232: will this actually take effect, since we have the following predicate:

https://github.com/stolostron/multicluster-global-hub/blob/c6b2b3f62f0083416af46ef0d96e1a5e86534fc8/agent/pkg/status/controller/policies/policy_status_sync.go#L42

yanmxa: It works. The local policy is synced by this controller.
https://github.com/stolostron/multicluster-global-hub/blob/c6b2b3f62f0083416af46ef0d96e1a5e86534fc8/agent/pkg/status/controller/local_policies/local_policy_sync.go#L34

clyang82: Can you help me understand why we add the finalizer for the local policy? I know that we add finalizer to help on the database cleanup if the policy is deleted. But we have watched such resources to help.

yanmxa: Well, I think that is a legacy left before. 

There are two main roles for the cleanup finalizer:
1. the global hub manager uses it to [clean up the data in the database when deleting this resource on the global hub](https://github.com/stolostron/multicluster-global-hub/blob/bc66f7a075a649aedc30fe7edca7ec858307f21d/manager/pkg/specsyncer/spec2db/controller/generic_spec_to_db_reconciler.go#L83-L84). 
2. the global hub operator uses it to clean up the resources before removing the MGH CR.

While I don't find the special role of the cleanup finalizer for the global hub agent. It will add a cleanup finalizer to all resources synced to the global hub, and removes the finalizers when deleting the objects from the regional hub.  
At the same time, when uninstalling the global hub, there is a pre-delete job to clean up all the global hub finalizers of the agent resources.

So we don't add finalziers to the local resources, which is the expected behavior.

yanmxa:
I need to add a test case to cover whether the local policy is deleted from the database without the finalizer.

agent/pkg/status/controller/generic/generic_status_sync_controller.go
yanmxa: @clyang82 Finalizer also plays an important role in the Global Hub Agent, the resources need the finalizer to remove the deleting obj from the bundle. So the deleted obj can be synced to the database.

morvencao (commented):
/lgtm

morvencao (commented):
/lgtm

Issue #343, opened 2023-03-13
Add a optional `ca.crt` for the global hub storage secret 


yanmxa:
According to the [document](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING), there are six connection modes, the following four types prefer, require, verify-ca and verify-full are more common.

- perfer, require can be used without a certificate
- verify-ca and verify-full require a certificate, the difference is that the former only verifies whether the certificate is issued by the postgres server ca or not, and the latter also verifies whether the hostname inside the certificate and the hostname in the URI are matched or not.

So we need to provide an optional ca.crt in addition to the necessary database_uri, and users can choose whether to provide this ca.crt certificate according to their configuration.

PR #345, opened 2023-03-14
enable postgres tls connection for global hub manager

yanmxa:
Resolved: https://github.com/stolostron/multicluster-global-hub/issues/343

### DIFF
--- a/manager/cmd/manager/main.go
+++ b/manager/cmd/manager/main.go
@@ -80,7 +80,9 @@ func parseFlags() (*managerconfig.ManagerConfig, error) {
 	pflag.DurationVar(&managerConfig.SyncerConfig.DeletedLabelsTrimmingInterval, "deleted-labels-trimming-interval",
 		5*time.Second, "The trimming interval of deleted labels.")
 	pflag.StringVar(&managerConfig.DatabaseConfig.ProcessDatabaseURL, "process-database-url", "",
-		"The URL of database server for the user.")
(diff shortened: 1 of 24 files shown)


manager/cmd/manager/main.go
```diff
@@ -80,9 +80,7 @@ func parseFlags() (*managerconfig.ManagerConfig, error) {
 	pflag.DurationVar(&managerConfig.SyncerConfig.DeletedLabelsTrimmingInterval, "deleted-labels-trimming-interval",
 		5*time.Second, "The trimming interval of deleted labels.")
 	pflag.StringVar(&managerConfig.DatabaseConfig.ProcessDatabaseURL, "process-database-url", "",
-		"The URL of database server for the process user.")
-	pflag.StringVar(&managerConfig.DatabaseConfig.TransportBridgeDatabaseURL,
-		"transport-bridge-database-url", "", "The URL of database server for the transport-bridge user.")
```
morvencao, line 85: This was not deleted when we refactored the code, because we may grant different permissions for DB syncers and transport bridge.
Anyway, UI think we may not need it in short term, OK to remove this.

yanmxa: Since the storage secret just provides one `database_uri`.  I remove this from the manager.
And yes, we can keep the `TransportBridgeDatabaseURL` in this PR, maybe we can delete it in the future.

morvencao (commented):
/lgtm

PR #346, opened 2023-03-15
disable the ACM Hub installing on agent by managedcluster label

yanmxa:


morvencao (commented):
will we support to dynamically change the label value, for example from `enabled` to `disabled`, and then uninstall the hub?

yanmxa:
> will we support to dynamically change the label value, for example from `enabled` to `disabled`, and then uninstall the hub?

@morvencao It supports dynamically updating the label value. When the value changed from `enabled` to `disabled`, it [won't uninstall the hub](https://github.com/stolostron/multicluster-global-hub/blob/c6b2b3f62f0083416af46ef0d96e1a5e86534fc8/operator/pkg/controllers/addon/manifests/templates/hubcluster/hubcluster-namespace.yaml#L10).

doc/reserved_names/README.md
```diff
@@ -7,6 +7,7 @@ Label | Description
 global-hub.open-cluster-management.io/managed-by=`global-hub-operator\|global-hub\|global-hub-agent` | If the value is `global-hub-operator`, it means the resources are created by the global hub operator. The global hub operator watches the resources based on this label.
 global-hub.open-cluster-management.io/local-resource= | This label is added during creating some resources. It is used to identify the resource is only applied to global hub cluster. It won't be transfered to the regional hub clusters.
 global-hub.open-cluster-management.io/agent-deploy-mode = `Hosted\| Default\| None` | This label is used on ManagedCluster.<br>`Hosted` means the HoH agent will be deployed on Hosting cluster.<br>`Default` means the HoH agent will be deployed on managed cluster.<br>`None` means the HoH agent should not be installed. 
+global-hub.open-cluster-management.io/hub-cluster-install = `Enabled\| Disabled` | This label is used on ManagedCluster.<br>`Enable` means the Global Hub addon will install ACM Hub on the agent cluster if there isn't found a hub.<br>`Disabled` means the Global Hub won't install ACM Hub on the agent and also as the default value for the label.
```
clyang82, line 10: I would expect if this label exists, then install ACM in the managed cluster. if without this label, then skip the installation. I do not care about the value.

clyang82:
Can you help to update `hub.open-cluster-management.io` with correct value? Thanks.

zhiweiyin318:
> > will we support to dynamically change the label value, for example from `enabled` to `disabled`, and then uninstall the hub?
> 
> @morvencao It supports dynamically updating the label value. When the value changed from `enabled` to `disabled`, it [won't uninstall the hub](https://github.com/stolostron/multicluster-global-hub/blob/c6b2b3f62f0083416af46ef0d96e1a5e86534fc8/operator/pkg/controllers/addon/manifests/templates/hubcluster/hubcluster-namespace.yaml#L10).

it looks weird if the label is disable but the hub has been installed and will not be removed. maybe it is better to only use the label without  value?

doc/reserved_names/README.md
yanmxa: Done!

yanmxa:
> > > will we support to dynamically change the label value, for example from `enabled` to `disabled`, and then uninstall the hub?
> > 
> > 
> > @morvencao It supports dynamically updating the label value. When the value changed from `enabled` to `disabled`, it [won't uninstall the hub](https://github.com/stolostron/multicluster-global-hub/blob/c6b2b3f62f0083416af46ef0d96e1a5e86534fc8/operator/pkg/controllers/addon/manifests/templates/hubcluster/hubcluster-namespace.yaml#L10).
> 
> it looks weird if the label is disable but the hub has been installed and will not be removed. maybe it is better to only use the label without value?

Yes, I removed the value.

PR #347, opened 2023-03-16
disable propagate the global policy/application to each regional hub cluster by GlobalResourceLabel

yanmxa:


manager/cmd/manager/main_test.go
```diff
@@ -150,6 +150,7 @@ func TestManager(t *testing.T) {
 					Namespace: constants.GHSystemNamespace,
 					Annotations: map[string]string{
 						constants.OriginOwnerReferenceAnnotation: "testing",
+						constants.GlobalHubGlobalResourceLabel:   "",
```
morvencao, line 153: Do we also need to add `global-hub.open-cluster-management.io/global-resource` annotation to configmap, so that it can be propagated to regional hubs?

yanmxa: No, It's a typo, and should in label.

morvencao (commented):
/lgtm

PR #361, opened 2023-03-23
expose image configs by mulitcluster-global-hub-operator environment

yanmxa:


operator/pkg/config/multiclusterglobalhub_config.go
```diff
@@ -107,6 +118,14 @@ func GetImageOverridesConfigmap(mgh *operatorv1alpha2.MulticlusterGlobalHub) str
...
 
 func SetImageOverrides(mgh *operatorv1alpha2.MulticlusterGlobalHub, cm *corev1.ConfigMap) error {
+	// image pull secret and policy
+	if mgh.Spec.ImagePullSecret != "" {
+		imageOverrides[ImagePullSecretKey] = mgh.Spec.ImagePullSecret
+	}
+	if mgh.Spec.ImagePullPolicy != "" {
+		imageOverrides[ImagePullPolicyKey] = string(mgh.Spec.ImagePullPolicy)
+	}
+
```
morvencao, line 128: it's weird to put image pull secret and pull policy in the image override map.

yanmxa: It seems that it is not so clear, and I changed it to get the pull policy and secret directly from mgh CR.

operator/pkg/controllers/addon/addon.go
```diff
@@ -190,6 +204,22 @@ func (a *HohAgentAddon) GetValues(cluster *clusterv1.ManagedCluster,
 		KlusterletWorkSA:       "klusterlet-work-sa",
 	}
 
+	if len(mgh.Spec.ImagePullSecret) > 0 {
```
clyang82, line 207: Get the `imagePullSecret` value from mch if not setting obviously. WDYT?

clyang82 (changes requested):

operator/pkg/controllers/addon/addon.go
yanmxa: @clyang82 
Of course, I think we can set the secret according to the priority.
1. The secret set by MCGH CR
2. Using the default secret `multiclusterhub-operator-pull-secret`
3. Using the user's default pull secret, that is to say both 1 and 2 are empty

operator/pkg/config/multiclusterglobalhub_config.go
```diff
@@ -125,38 +136,49 @@ func SetImageOverrides(mgh *operatorv1alpha2.MulticlusterGlobalHub, cm *corev1.C
...
-
 	return nil
 }
 
 // GetImage is used to retrieve image for given component
 func GetImage(componentName string) string {
 	return imageOverrides[componentName]
 }
+
+// GetImagePullSecret returns the image pull secret name and data
```
clyang82, line 147: I think it is a little complex here. You have designed this func `GetImagePullSecret` to return the imagepullsecret name and imagepullsecret data. Actually, we only need to retrieve the imagepullsecret to create in mc side by using manifestwork. we do not need to call this func in deploying grafana, manager, etc.

I think the ideal way is to get the imagepullsecret name from the mcgh always. if user sets the imagepullsecret, then use it. if user does not set the imagepullsecret, we set the default value. in other words, you can always get the correct imagepullsecret name from mcgh.

We only need to retrieve the imagepullsecret secret when the agent is enabled. the previous code can be reusable. hope it helps.

clyang82 (changes requested):

operator/pkg/config/multiclusterglobalhub_config.go
yanmxa: This function is used to validate whether the secrets exist. Since you mean that the manager and grafana don't need to verify the secret has been created. I just keep this validation in the addon.

clyang82 (changes requested):
Please add imagePullSecret in `mch` https://github.com/stolostron/multicluster-global-hub/blob/2691a03a03a2fcefd82c7c184eec06a67ae2d1c2/operator/pkg/controllers/addon/manifests/templates/hubcluster/hubcluster-mch.yaml#L13  It is used to use this imagepullsecret to download acm images.

yanmxa:
> 

Done.

operator/pkg/config/multiclusterglobalhub_config.go
```diff
@@ -125,38 +136,49 @@ func SetImageOverrides(mgh *operatorv1alpha2.MulticlusterGlobalHub, cm *corev1.C
...
+	// 1. get image pull secret from mgh
+	if mgh != nil && len(mgh.Spec.ImagePullSecret) > 0 {
+		imagePullSecret := &corev1.Secret{}
+		err := c.Get(ctx, types.NamespacedName{
+			Namespace: mgh.GetNamespace(),
+			Name:      mgh.Spec.ImagePullSecret,
+		}, imagePullSecret, &client.GetOptions{})
+		switch {
+		case err == nil:
+			imagePullSecretName = operatorconstants.DefaultImagePullSecretName
```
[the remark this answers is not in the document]
yanmxa: This should be `imagePullSecret.GetName()`. the default image is an error.

clyang82 (approved):
/lgtm

PR #365, opened 2023-03-28
remove global hub finalizer from local resources and managedcluster.

yanmxa:
refs: https://issues.redhat.com/secure/RapidBoard.jspa?rapidView=16808&quickFilter=105908
1. We used the `finalizer` to delete obj from the bundles before, and now we have to depend on the `not found error` to remove the bundle obj so that can be deleted from the database.
2. Since there is no `UID` for the deleted obj, I have to add a new `Namespacedname` to the bundle object for retrieval.
refs: https://issues.redhat.com/browse/ACM-4597
fixed: https://github.com/stolostron/multicluster-global-hub/issues/320


clyang82 (approved):
LGTM

PR #380, opened 2023-04-06
Event exporter

clyang82:
ref: https://issues.redhat.com/browse/ACM-4099


pkg/testdata/event/kube-event-exporter-bad-config.yaml
yanmxa, line 1: Why do we need this bad event exporter config?

clyang82: for UT purpose, refer to https://github.com/stolostron/multicluster-global-hub/blob/def309a0494131094307ea5399f029b5e086f94f/agent/cmd/agent/main_test.go#L353

Issue #317, opened 2023-02-13
Add another regional hub for the E2E tests


yanmxa:
The previous E2E Global Hub environment is composed of a regional hub and two managed clusters. That may cause some test cases not to be covered. Therefore, building two regional hubs and each with a managed cluster for the E2E tests is necessary.
![image](https://user-images.githubusercontent.com/19286664/218356794-2ddaa9eb-701a-428d-ab73-cbdc835a82af.png)

PR #385, opened 2023-04-09
Fix e2e tests for another localhub

MMMsc:
1. fix e2e_setup for another localhub
2. fix e2e_tests
3. add e2e_tests_validation

fixed: https://github.com/stolostron/multicluster-global-hub/issues/317


cicd-scripts/run-prow-e2e-test.sh
```diff
@@ -23,6 +23,7 @@ echo "export VERBOSE=6" >> ${ROOT_DIR}/test/resources/env.list
 scp "${OPT[@]}" -r ../multicluster-global-hub "$HOST:$HOST_DIR"
 
 ssh "${OPT[@]}" "$HOST" sudo yum install gcc git wget jq -y 
+ssh "${OPT[@]}" "$HOST" "sudo sh -c 'echo \"fs.inotify.max_user_watches=524288\" >> /etc/sysctl.conf && echo \"fs.inotify.max_user_instances=1024\" >> /etc/sysctl.conf && sysctl -p /etc/sysctl.conf'"
```
yanmxa, line 26: add a comment on why we need this config.

test/pkg/e2e/hoh-local-policy_test.go
```diff
@@ -37,47 +40,58 @@ const (
...
+						TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+				}
+				httpClient = &http.Client{Timeout: time.Second * 20, Transport: transport}
+				managedClusters, err = getManagedCluster(httpClient, httpToken)
 				if err != nil {
 					return err
 				}
-				managedClusterUID1 = string(managedClusters[0].GetUID())
-				managedClusterName1 = managedClusters[0].Name
+				if len(managedClusters) == 0 {
```
yanmxa, line 60: I think you need to make sure how many managed cluster which is expected, instead of determining whether it is empty.

test/pkg/e2e/hoh-local-policy_test.go
```diff
@@ -37,47 +40,58 @@ const (
...
+					return fmt.Errorf("managed cluster is not exist")
+				}
 				return nil
-			}, 1*time.Minute, 1*time.Second).ShouldNot(HaveOccurred())
+			}, 3*time.Minute, 5*time.Second).ShouldNot(HaveOccurred())
```
yanmxa, line 64: Why does it take us 3 minutes to get the managed cluster, is it a bit too long?

test/pkg/e2e/hoh-local-policy_test.go
```diff
@@ -37,47 +40,58 @@ const (
...
-			leafhubClient, err = clients.ControllerRuntimeClient(clients.LeafHubClusterName(), scheme)
-			Expect(err).Should(Succeed())
-
-			By("Create local policy namespace on leafhub")
-			err = leafhubClient.Create(context.TODO(), &v1.Namespace{ObjectMeta: metav1.ObjectMeta{
-				Name: LOCAL_POLICY_NAMESPACE,
-			}}, &client.CreateOptions{})
-			if !errors.IsAlreadyExists(err) {
+			// get multiple leafhubs
+			leafhubNames = clients.GetLeafHubClusterNames()
```
yanmxa, line 74: We need to determine how many leaf hubs we expect to get.

test/pkg/e2e/hoh-local-policy_test.go
```diff
@@ -87,21 +101,22 @@ var _ = Describe("Apply local policy to the managed clusters", Ordered,
...
+				for _, managedCluster := range managedClusters {
+					err := updateClusterLabel(httpClient, patches, httpToken, string(managedCluster.UID))
+					if err != nil {
+						return err
+					}
+					managedClusterInfo, err := getManagedClusterByName(httpClient, httpToken, managedCluster.Name)
+					if err != nil {
+						return err
+					}
+					if val, ok := managedClusterInfo.Labels[LOCAL_POLICY_LABEL_KEY]; ok {
```
yanmxa, line 113: What if there is no the such label on the managed cluster? The logic is not correct.

test/pkg/e2e/hoh-local-policy_test.go
MMMsc: Considering add a function ManagedClusterNumber in client interface

test/pkg/e2e/hoh-local-policy_test.go
MMMsc: Synchronize with several other e2e.

test/pkg/e2e/hoh-local-policy_test.go
MMMsc: Considering verify the number of the leafhubs mentioned in options-local.yaml

test/pkg/e2e/hoh-local-policy_test.go
MMMsc: Is it too redundant to check if it is empty first and then check the logic of the label's judgment statement?

test/pkg/e2e/hoh-local-policy_test.go
```diff
@@ -164,29 +184,36 @@ var _ = Describe("Apply local policy to the managed clusters", Ordered,
...
-						if err := rows.Scan(policy); err != nil {
+						policy := &policiesv1.Policy{}
+						leafhub := ""
+						if err := rows.Scan(&leafhub, policy); err != nil {
 							return err
 						}
 						fmt.Printf("local_spec.policies: %s/%s \n", policy.Namespace, policy.Name)
-						if policy.Name == LOCAL_POLICY_NAME && policy.Namespace == LOCAL_POLICY_NAMESPACE {
-							return nil
+						if policy.Name != LOCAL_POLICY_NAME || policy.Namespace != LOCAL_POLICY_NAMESPACE {
```
yanmxa, line 208: verify the leafhub is contained in each leafhubnames.

test/pkg/e2e/hoh-local-policy_test.go
```diff
@@ -359,7 +405,14 @@ var _ = Describe("Apply local policy to the managed clusters", Ordered,
 						if err := rows.Scan(&policyId, &cluster, &leafhub); err != nil {
 							return err
 						}
-						if cluster == managedClusterName1 && leafhub == clients.LeafHubClusterName() {
```
yanmxa, line 362: remove the foundPolicy flag.

test/pkg/e2e/hoh-validation_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	Context("Check all the parameters for e2e-tests", func() {	
+		var HubClusterNum = 2
+		var ManagedClusterNum = 1
+		It("Check the num of hub clusters and managed clusters on options-local.yaml", func() {
+			opt := testOptions.ManagedClusters
+			var hubClusters []string
+			var managedClusters []string
+
+			Eventually(func() error {
```
yanmxa, line 78: I don't think we need `Eventually` in the optional validation. Because the `optional.yaml` has been generated and will not be updated when running this test spec.

test/pkg/e2e/hoh-validation_test.go
```diff
@@ -0,0 +1,114 @@
...
+					} else {
+						managedClusters = append(managedClusters, c.Name)
+					}
+				}
+				
+				if len(hubClusters) != HubClusterNum || len(managedClusters) != HubClusterNum*ManagedClusterNum {
+					return fmt.Errorf("generate %d hub cluster and %d managed cluster error", HubClusterNum, ManagedClusterNum)
+				}
+				return nil
+			}, 3*time.Minute, 5*time.Second).ShouldNot(HaveOccurred())
```
yanmxa, line 91: It won't take 3 minutes to verify the static file `option.yaml`, suggest to reduce this timeout.

test/pkg/utils/client.go
```diff
@@ -24,6 +24,9 @@ type Client interface {
 	RestConfig(clusterName string) (*rest.Config, error)
 	HubClusterName() string
```
yanmxa, line 25: Maybe you need to remove line 25 ~ 29, cause these variables do not belong to the `Client`.

Instead, you can define the global variables and initialize them before running the suites, such as: `GlobalHubName`, `LeafHubNames`. If the `LeafHubClusterNumber()` and the `ManagedClusterNumber() ` are just used in the validation spec, you can move the code to `hoh-validatation_test.go` directly.

test/pkg/e2e/hoh-local-policy_test.go
yanmxa: @MMMsc I mean you need to reduce the timeout.

test/pkg/e2e/hoh-local-policy_test.go
yanmxa: The ManagedClusterNumber does not belong to `Client`. Just define the global variable like `ExepectedManagedCLusterNum` and verify whether it is consistent with the `option-local.yaml` in the validation spec.

test/pkg/e2e/hoh-local-policy_test.go
yanmxa: @MMMsc Same as above.

test/pkg/e2e/hoh-local-policy_test.go
MMMsc: The leafhub number and the managedcluster number is being predefined in the validation_test line 71. Therefore, I think it is reasonable to verify the number of clusters in the Options here.

test/pkg/e2e/hoh-local-policy_test.go
MMMsc: completed

test/pkg/e2e/hoh-local-policy_test.go
MMMsc: completed

test/pkg/e2e/hoh-validation_test.go
MMMsc: completed

test/pkg/e2e/hoh-validation_test.go
MMMsc: completed

clyang82 (approved):
Overall looks good for me. Thanks @MMMsc for your contribution.

We can have the following PRs to address the comments. Need to think over the new requirements that the tests can be running in canary tests. the basic requirement is the ACM/OCM cluster is passed via configuration file. We need to write golang to setup the global hub environment.

Leave @yanmxa to add `lgtm` if he is fine with the current changes.

PR #392, opened 2023-04-17
fixed the e2e issue that managed cluster label isn't added

yanmxa:
Description: The e2e test might fail sometimes. Because adding labels one by one in a for loop can easily cause timeouts. So instead, we add labels to each cluster first, and then check if the labels are added, so that we can add labels to multiple clusters in parallel and finish the test within the specified time.


yanmxa:
> Can you update the reason in the description?

Done!

PR #396, opened 2023-04-19
force delete the global hub addons when waiting for the deletion timeout

yanmxa:
Resolves: https://issues.redhat.com/browse/ACM-5029

- Add force deletion of the global hub addons 
- Prune job only removes the global hub finalizer from the global resources

### DIFF
--- a/operator/pkg/controllers/hubofhubs/globalhub_prune.go
+++ b/operator/pkg/controllers/hubofhubs/globalhub_prune.go
@@ -35,7 +35,7 @@ func (r *MulticlusterGlobalHubReconciler) pruneGlobalHubResources(ctx context.Co
 	log.Info("deleted ClusterManagementAddon", "name", operatorconstants.GHClusterManagementAddonName)
 
 	// prune the hub resources until all addons are cleaned up
-	if err := r.waitUtilAddonDeleted(ctx, log); err != nil {
+	if err := r.waitUtilAddonDeleted(ctx, log, 5*time.Minute); err != nil {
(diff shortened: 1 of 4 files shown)


operator/pkg/controllers/hubofhubs/globalhub_prune.go
```diff
@@ -166,27 +166,82 @@ func (r *MulticlusterGlobalHubReconciler) deleteClusterManagementAddon(ctx conte
 	return nil
 }
 
-func (r *MulticlusterGlobalHubReconciler) waitUtilAddonDeleted(ctx context.Context, log logr.Logger) error {
-	ctx, cancel := context.WithTimeout(ctx, 60*time.Second)
+// delete the managedclusteraddons with graceful duration, if reach out of the graceful duration,
+// then force delete them.
+func (r *MulticlusterGlobalHubReconciler) waitUtilAddonDeleted(ctx context.Context, log logr.Logger,
```
clyang82, line 171: I remembered addon framework provide this capability. Do we really need to handle the force deletion in our logic?

clyang82: @zhiweiyin318 ^

zhiweiyin318: import-controller will force delete all addon CRs and manifestWorks after the cluster is unavailable.

yanmxa:
Close this PR Cause we don't need to add the force deletion, instead to diagnose the reason why it can't be deleted.

PR #403, opened 2023-04-24
add cronjob to sync data to local_status.compliance_history

yanmxa:
Reference: https://issues.redhat.com/browse/ACM-4099

### DIFF
--- a/manager/cmd/manager/main.go
+++ b/manager/cmd/manager/main.go
@@ -25,6 +25,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
 	managerconfig "github.com/stolostron/multicluster-global-hub/manager/pkg/config"
+	"github.com/stolostron/multicluster-global-hub/manager/pkg/cronjob"
 	"github.com/stolostron/multicluster-global-hub/manager/pkg/nonk8sapi"
 	"github.com/stolostron/multicluster-global-hub/manager/pkg/scheme"
 	"github.com/stolostron/multicluster-global-hub/manager/pkg/specsyncer/db2transport/db/postgresql"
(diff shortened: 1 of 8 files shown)


manager/cmd/manager/main.go
```diff
@@ -256,8 +257,15 @@ func doMain(ctx context.Context, restConfig *rest.Config) int {
...
-	log.Info("Starting the Cmd.")
+	log.Info("starting the cron job scheduler")
+	jobScheduler, err := cronjob.StartJobScheduler(ctx, processPostgreSQL.GetConn())
+	if err != nil {
+		log.Error(err, "failed to start job scheduler")
+		return 1
+	}
+	defer jobScheduler.Stop()
 
+	log.Info("Starting the Manager")
```
morvencao, line 268: will this conflict with database init, because we init the database schemas in controller, the cronjob may run before the database init?

yanmxa: @morvencao Initialize database is located in the global hub operator, and the cronjob is running in the manager. So the job will always run after the database init.

manager/pkg/cronjob/sheduler.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/go-co-op/gocron"
+	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/stolostron/multicluster-global-hub/manager/pkg/cronjob/task"
+)
+
+func StartJobScheduler(ctx context.Context, pool *pgxpool.Pool) (*gocron.Scheduler, error) {
+	log := ctrl.Log.WithName("cronjob-scheduler")
+	s := gocron.NewScheduler(time.UTC)
+
+	complianceJob, err := s.Every(1).Day().At("01:00").Tag("LocalCompliance").DoWithJobDetails(
```
clyang82, line 18: why not `00:00?

clyang82: may be `LocalComplianceHistory` for readable.

manager/pkg/cronjob/task/compliance_history.go
```diff
@@ -0,0 +1,68 @@
...
+
+	conn, err := pool.Acquire(ctx)
+	if err != nil {
+		return err
+	}
+	defer conn.Release()
+
+	// clean the recent data  HOURS/MINUTES/SECONDS
+	result, err := conn.Exec(ctx,
+		`DELETE FROM local_status.compliance_history WHERE updated_at BETWEEN NOW() - INTERVAL '1 HOURS' AND NOW()`)
```
clyang82, line 28: why need drop records?

manager/pkg/cronjob/sheduler.go
yanmxa: Trigger the time at `00:00` and update the tag with `LocalComplianceHistory `.

manager/pkg/cronjob/task/compliance_history.go
yanmxa: remove the clean action be insert data.

PR #427, opened 2023-05-12
add event export to controller manager

yanmxa:
- Add event export to controller manager to support HA
- Convert the event export initialization and start() to a goroutine controller, so that it won't block other processes when it failed.


clyang82 (approved):
LGTM

PR #430, opened 2023-05-15
update the TLS configration for event exporter

yanmxa:
There are two cases for event exporter connection:

- production environment: we need to enable the TLS.
    - the /kafka-certs/client.crt and /kafka-certs/client.key file are valid.
    - insecureSkipVerify is false
- e2e test environment: since the e2e pod id(hostname) can't be signed by the client ca, we need to disable the TLS.
    - the /kafka-certs/client.crt and /kafka-certs/client.key files exist but are empty. Which might cause the [Kafka sinker initialize](https://github.com/resmoio/kubernetes-event-exporter/blob/f4b7ad969e5c78fd0538d5807ac3363066f8b17c/pkg/sinks/kafka.go#L156) error, and it will be fixed once this [PR](https://github.com/resmoio/kubernetes-event-exporter/pull/80) is merged.
    - insecureSkipVerify is true.

### DIFF
--- a/agent/pkg/controllers/clusterclaim_controllers_test.go
+++ b/agent/pkg/controllers/clusterclaim_controllers_test.go
@@ -25,7 +25,7 @@ const (
 	MCHVersion = "2.6.0"
 )
 
-var _ = Describe("controller", Ordered, func() {
+var _ = Describe("claim controllers", Ordered, func() {
 	BeforeEach(func() {
 		cleanup(ctx, mgr.GetClient())
 	})
--- a/agent/pkg/controllers/controllers_suite_test.go
+++ b/agent/pkg/controllers/controllers_suite_test.go
@@ -11,7 +11,6 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	"github.com/stolostron/multicluster-global-hub/agent/pkg/controllers"
 	mchv1 "github.com/stolostron/multiclusterhub-operator/api/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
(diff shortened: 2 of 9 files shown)


clyang82:
I do not fully understand the fix. the issue reports ` unknown flag: --kafka-ca-path`, but your fix is to skip the kafka tls for event exporter.

yanmxa:
> I do not fully understand the fix. the issue reports ` unknown flag: --kafka-ca-path`, but your fix is to skip the kafka tls for event exporter.

`unknown flag: --kafka-ca-path` is not an issue, you just need to using the latest image and the error will disapear.

yanmxa:
There are two cases for event exporter connection:
1. production environment: we need to enable the TLS.
    - the `/kafka-certs/client.crt` and `/kafka-certs/client.key` file are valid. 
    - `insecureSkipVerify` is false
2. e2e test environment: since the e2e pod id(hostname) can't be signed by the client ca, we need to disable the TLS.
    - the `/kafka-certs/client.crt` and `/kafka-certs/client.key` files exist but are empty. Which might cause  the [Kafka sinker initialize](https://github.com/resmoio/kubernetes-event-exporter/blob/f4b7ad969e5c78fd0538d5807ac3363066f8b17c/pkg/sinks/kafka.go#L156) error, and it will be fixed once this [PR](https://github.com/resmoio/kubernetes-event-exporter/pull/80) is merged.
    - `insecureSkipVerify` is true.

yanmxa:
> that means https://issues.redhat.com/browse/ACM-5465 is an environment issue? if so, we just need to append it in that issue. should not refer this PR to that issue.

Yes.

clyang82:
The code looks good for me.

yanmxa:
> Good. please update the description to remove the refs and have correct information for this PR.

Done.

PR #434, opened 2023-05-16
Fix cluster_id is NULL in local_status.compliance table sometimes

clyang82:
there is a case that the cluster_id in status.managed_clusters table is empty before insert the data into local_status.compliance table. so create trigger before update to ensure the cluster_id is inserted.

### DIFF
--- a/operator/pkg/controllers/hubofhubs/database/4.trigger.sql
+++ b/operator/pkg/controllers/hubofhubs/database/4.trigger.sql
@@ -69,8 +69,8 @@ DROP TRIGGER IF EXISTS set_timestamp ON spec.subscriptions;
 CREATE TRIGGER set_timestamp BEFORE UPDATE ON spec.subscriptions FOR EACH ROW EXECUTE FUNCTION public.trigger_set_timestamp();
 
 DROP TRIGGER IF EXISTS update_compliance_table ON local_status.compliance;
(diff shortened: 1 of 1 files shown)


operator/pkg/controllers/hubofhubs/database/4.trigger.sql
```diff
@@ -70,5 +70,7 @@ CREATE TRIGGER set_timestamp BEFORE UPDATE ON spec.subscriptions FOR EACH ROW EX
 
 DROP TRIGGER IF EXISTS update_compliance_table ON local_status.compliance;
 CREATE TRIGGER update_compliance_table AFTER INSERT ON local_status.compliance FOR EACH ROW EXECUTE FUNCTION public.set_cluster_id_to_local_compliance();
+CREATE TRIGGER update_compliance_table BEFORE UPDATE ON local_status.compliance FOR EACH ROW EXECUTE FUNCTION public.set_cluster_id_to_local_compliance();
 DROP TRIGGER IF EXISTS update_compliance_table ON status.compliance;
 CREATE TRIGGER update_compliance_table AFTER INSERT ON status.compliance FOR EACH ROW EXECUTE FUNCTION public.set_cluster_id_to_compliance();
+CREATE TRIGGER update_compliance_table BEFORE UPDATE ON local_status.compliance FOR EACH ROW EXECUTE FUNCTION public.set_cluster_id_to_compliance();
```
yanmxa, line 76: typo `status.compliance` rather than `local_status.compliance`

PR #435, opened 2023-05-16
remove relatedImages from CSV

yanmxa:
Since the `operator-sdk` hasn't have the feature to remove `relatedImages` field when generate bundle/manifests. Here we just manually remove it from bundle CSV file.


clyang82 (approved):
LGTM

PR #436, opened 2023-05-16
[ACM-5468] Convert dashboards to use UTC timezone

dislbenn:
Jira Issue: https://issues.redhat.com/browse/ACM-5468

### DIFF
--- a/operator/pkg/controllers/hubofhubs/manifests/grafana/acm-global-cluster-group-compliancy-overview.yaml
+++ b/operator/pkg/controllers/hubofhubs/manifests/grafana/acm-global-cluster-group-compliancy-overview.yaml
@@ -24,8 +24,7 @@ data:
       "editable": true,
       "gnetId": null,
       "graphTooltip": 0,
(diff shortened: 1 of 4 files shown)


clyang82 (approved):
LGTM

clyang82:
I looked at the dashboards in observability, the timezone is empty or browser. set as `utc` maybe confusion. I am thinking over whether we add start_from and end_from to help on.

dislbenn:
@clyang82 It seems like the `utc` timezone will only be needed for dashboards that are using the `Postgres` data source. The Observability dashboards are using the timezone based on the user's local browser.

clyang82:
> 

If you check the global hub dashboards, you can see they are using the timezone based on the user's local browser as well.

dislbenn:
After discussion on Slack, I will close this PR to allow dashboards to remain consistent with the user's local browser timezone. We will need to document this to avoid confusion in cases where the user would expect the data to be plotted at midnight or `00:00:00`.

PR #439, opened 2023-05-17
remove the deprecated `io/ioutil` package

yanmxa:

Resolve the following issue when building the manager image:
```bash
SA1019: "io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details. (staticcheck)
        "io/ioutil"
        ^
```


clyang82:
Good catch. But it seems you need to ensure the code coverage is greater than 70% so that it can be merged.

PR #431, opened 2023-05-15
customize scheduler interval.

morvencao:
This PR tries to support customize scheduler interval for moving policy compliance to history, can be 'month', 'week', 'day', 'hour', 'minute' or 'second'.

PR #440, opened 2023-05-17
enable data simulation for local policy compliance history.

morvencao:
followup PR for https://github.com/stolostron/multicluster-global-hub/pull/431 to enable data simulation for local policy compliance history.

### DIFF
--- a/doc/simulation/local-policies/README.md
+++ b/doc/simulation/local-policies/README.md
@@ -23,6 +23,7 @@ kind: MulticlusterGlobalHub
 metadata:
   annotations:
     mgh-scheduler-interval: minute # change scheduler interval of moving to compliance_history to 1 minute
+    mgh-enable-simulation: "true" # enable data simulation for scheduler job
   name: multiclusterglobalhub
   namespace: open-cluster-management
 spec:
--- a/doc/simulation/local-policies/README.md
+++ b/doc/simulation/local-policies/README.md
@@ -23,7 +23,6 @@ kind: MulticlusterGlobalHub
 metadata:
   annotations:
     mgh-scheduler-interval: minute # change scheduler interval of moving to compliance_history to 1 minute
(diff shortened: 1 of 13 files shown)


doc/simulation/local-policies/README.md
```diff
@@ -23,6 +23,7 @@ kind: MulticlusterGlobalHub
 metadata:
   annotations:
     mgh-scheduler-interval: minute # change scheduler interval of moving to compliance_history to 1 minute
+    mgh-enable-simulation: "true" # enable data simulation for scheduler job
```
clyang82, line 26: I am not a fan to introduce another new annotation for this purpose. since it is only for internal testing, you can just check if mgh-scheduler-interval exists, then it is for data simulation.  does it make sense?

morvencao: done.

manager/pkg/cronjob/task/compliance_history.go
```diff
@@ -121,23 +121,36 @@ func insertToLocalComplianceHistory(ctx context.Context, pool *pgxpool.Pool, vie
...
+		if err != nil {
+			fmt.Printf("exec failed: %v, retrying\n", err)
 			return false, nil
 		}
 		if insertError = tx.Commit(ctx); insertError != nil {
 			fmt.Printf("commit failed: %v, retrying\n", insertError)
 			return false, nil
 		} else {
+			if enableSimulation {
+				simulationCounter = simulationCounter + 1
```
clyang82, line 152: simulationCounter++ should be in `syncToLocalComplianceHistory` method. if you put here, it leads to missing data in a day.

morvencao: nice catch!

clyang82 (approved):
LGTM

PR #446, opened 2023-05-19
add the `cluster_id/root_policy_id/compliance` for the exported events

yanmxa:
Refs: https://issues.redhat.com/browse/ACM-4099
The enhanced event will like this:
```json
{
  "metadata":{
    "name":"local-placement.policy-limitrange.1760747c021c8f0c",
    "namespace":"kind-hub1-cluster1",
    "uid":"6519fdb1-4cc9-4c81-a5d7-9e47210ab4c3",
    "resourceVersion":"930321",
    "creationTimestamp":"2023-05-19T05:32:16Z"
  },
  "reason":"PolicyStatusSync",
  "message":"Policy local-placement.policy-limitrange status was updated in cluster namespace kind-hub1-cluster1",
  "source":{
    "component":"policy-status-sync"
  },
  "firstTimestamp":"2023-05-19T05:32:16Z",
  "lastTimestamp":"2023-05-19T05:32:16Z",
  "count":1,
  "type":"Normal",
  "eventTime":null,
  "reportingComponent":"",
  "reportingInstance":"",
  "clusterName":"kind-hub1",
  "involvedObject":{
    "kind":"Policy",
    "namespace":"kind-hub1-cluster1",
    "name":"local-placement.policy-limitrange",
    "uid":"36e404d1-f138-46cc-99e0-a086d816f83a",
    "apiVersion":"policy.open-cluster-management.io/v1",
    "resourceVersion":"930320",
    "labels":{
      "policy.open-cluster-management.io/cluster-compliance":"NonCompliant",
      "policy.open-cluster-management.io/cluster-id":"d8328442-65d1-48a3-9bdb-eafbd73f28d4",
      "policy.open-cluster-management.io/cluster-name":"kind-hub1-cluster1",
      "policy.open-cluster-management.io/cluster-namespace":"kind-hub1-cluster1",
      "policy.open-cluster-management.io/root-policy":"local-placement.policy-limitrange",
      "policy.open-cluster-management.io/root-policy-id":"0316ca75-a3fd-439b-95fb-6e628af30792"
    },
    "annotations":{
      "argocd.argoproj.io/compare-options":"IgnoreExtraneous",
      "policy.open-cluster-management.io/categories":"PR.IP Information Protection Processes and Procedures",
      "policy.open-cluster-management.io/controls":"PR.IP-1 Baseline Configuration",
      "policy.open-cluster-management.io/standards":"NIST-CSF"
    }
  }
}
```


clyang82:
It can be expensive to retrieve policy and managedcluster for each event. I am looking for adding such information when we emit the events in the policy side. but it seems we can not have such information from the existing mechanism, we have to retrieve in policy side if we add them into the original event.

clyang82:
in the agent side, we already watch the managedcluster and policy. Can we do a cache instead of retrieving it again?

yanmxa:
> It can be expensive to retrieve policy and managedcluster for each event. I am looking for adding such information when we emit the events in the policy side. but it seems we can not have such information from the existing mechanism, we have to retrieve in policy side if we add them into the original event.
1. It only retrieves the events which aren't from the root policy
2. We can cache the cluster_id and policy_id, but we can't cache the  policy compliance cause it changes over time
3. A good news is that the client from the controller runtime manager that [will access the k8s apiserver for write operations and the cache for read operations.](https://github.com/kubernetes-sigs/controller-runtime/blob/d18941971dff2fe91e1eca8fa0ee25f8cfd0ab5e/pkg/manager/client_builder.go#L49-L61) .

yanmxa:
Cause we only read the policy and managedcluster when we emit the events. I think we don't need to worry about the request load for apiserver.

clyang82:
`policy.open-cluster-management.io/cluster-compliance`: we may not put it into labels, because the compliance status can be mismatched and we will follow a simple logic for the compliance status (https://issues.redhat.com/browse/ACM-5453)

yanmxa:
> `policy.open-cluster-management.io/cluster-compliance`: we may not put it into labels, because the compliance status can be mismatched and we will follow a simple logic for the compliance status (https://issues.redhat.com/browse/ACM-5453)

@clyang82 I read the logic you posted here. I think the `compliance` from the event is the original data. We use it to calculate [compliance_changed_frequency](https://github.com/stolostron/multicluster-global-hub/blob/1acda384d759a6ee15906521916f76dbe0cfff58/operator/pkg/controllers/hubofhubs/database/3.tables.sql#L114) and based on the **compliance_changed_frequency** to determine the [compliance](https://github.com/stolostron/multicluster-global-hub/blob/1acda384d759a6ee15906521916f76dbe0cfff58/operator/pkg/controllers/hubofhubs/database/3.tables.sql#L114).

So I think it is normal that the compliance status in events is different from the status calculated by history changed frequency, because the former changes with time, and the latter is a statistical value based on the change of the former.

clyang82 (approved):
LGTM

PR #447, opened 2023-05-23
fix empty schedule interval.

morvencao:
ref: https://issues.redhat.com/browse/ACM-5619
if `SchedulerInterval` is not specified in MGH instance, then don't set `--scheduler-interval` to manager.


clyang82 (approved):
LGTM

PR #450, opened 2023-05-24
disable grc in global hub.

morvencao:
ref: https://issues.redhat.com/browse/ACM-5632


operator/pkg/controllers/hubofhubs/globalhub_mch.go
```diff
@@ -0,0 +1,103 @@
...
+				},
+			},
+		}
+		grcDisabledByMGH = true
+	}
+
+	return mch, grcDisabledByMGH
+}
+
+func getMCH(ctx context.Context, k8sClient client.Client) (*mchv1.MultiClusterHub, error) {
```
yanmxa, line 85: There has a [function](https://github.com/stolostron/multicluster-global-hub/blob/a0f3a47d4a3dd70d10ec9424e7ee8aa0ef79d531/agent/pkg/controllers/utils.go#L19) to getMCH has been implemented before. Maybe we can just keep one of these to the `pkg`.

PR #454, opened 2023-05-30
backup managed cluster to history after deleting it

yanmxa:
Refers: https://issues.redhat.com/browse/ACM-5521

### DIFF
--- a/operator/pkg/controllers/hubofhubs/database/4.trigger.sql
+++ b/operator/pkg/controllers/hubofhubs/database/4.trigger.sql
@@ -74,6 +74,7 @@ DROP TRIGGER IF EXISTS update_compliance_table ON status.compliance;
 CREATE TRIGGER update_compliance_table AFTER INSERT OR UPDATE ON status.compliance FOR EACH ROW WHEN (pg_trigger_depth() < 1) EXECUTE FUNCTION public.set_cluster_id_to_compliance();
 
 --- history tiger
+DROP TRIGGER IF EXISTS trigger_backup_deleted_managed_cluster ON status.managed_clusters;
 CREATE TRIGGER trigger_backup_deleted_managed_cluster
 AFTER DELETE ON status.managed_clusters
 FOR EACH ROW


hchenxa:
@yanmxa , do you think we also need to update the grafana dashboard to select the data from this table history.managed_clusters as well?

yanmxa:
> @yanmxa , do you think we also need to update the grafana dashboard to select the data from this table history.managed_clusters as well?

Yes, I think maybe we need to combine the history and status together for all of the manage_clusters.
```pg
SELECT leaf_hub_name, cluster_name, cluster_id, payload, error
FROM status.managed_clusters
UNION ALL
SELECT leaf_hub_name, cluster_name, cluster_id, payload, error
FROM history.managed_clusters;
```

operator/pkg/controllers/hubofhubs/database/3.tables.sql
```diff
@@ -1,83 +1,110 @@
...
+    CONSTRAINT local_policies_unique_constraint UNIQUE (id, cluster_id, compliance_date)
+);
+
+CREATE TABLE IF NOT EXISTS history.local_compliance_job_log (
+    name varchar(63) NOT NULL,
+    start_at timestamp NOT NULL DEFAULT now(),
+    end_at timestamp NOT NULL DEFAULT now(),
+    total int8,
+    inserted int8,
+    offsets int8,
```
clyang82, line 96: what is changed comparing the previous one?

operator/pkg/controllers/hubofhubs/database/3.tables.sql
```diff
@@ -293,90 +301,6 @@ CREATE TABLE IF NOT EXISTS event.local_policies (
     CONSTRAINT local_policies_unique_constraint UNIQUE (policy_id, cluster_id, created_at)
 );
 
-ALTER TABLE history.applications DROP CONSTRAINT IF EXISTS applications_pkey;
```
clyang82, line 296: why remove those?

yanmxa: This way of writing is the same as the following way of writing, the deletion here just makes it more concise.
```pg
    id uuid PRIMARY KEY,
```

operator/pkg/controllers/hubofhubs/database/3.tables.sql
yanmxa: Nothing changed. I just move the history table together.

PR #457, opened 2023-05-30
skip reconcile database when all the tables have been initialized

yanmxa:
avoid executing all SQL statements to reconcile the database when all the tables have been initialized.

### DIFF
--- a/operator/pkg/controllers/hubofhubs/globalhub_database.go
+++ b/operator/pkg/controllers/hubofhubs/globalhub_database.go
@@ -5,13 +5,28 @@ import (
 	"embed"
 	"fmt"
 	iofs "io/fs"
+	"strings"
+	"time"
 
+	"github.com/jackc/pgx/v4"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	operatorv1alpha2 "github.com/stolostron/multicluster-global-hub/operator/apis/v1alpha2"
 	"github.com/stolostron/multicluster-global-hub/operator/pkg/condition"
 	"github.com/stolostron/multicluster-global-hub/operator/pkg/config"
 	"github.com/stolostron/multicluster-global-hub/pkg/database"
+)
+
+var (
+	CheckDatabaseInterval = time.Duration(10) * time.Second
+	SchemeTableCounter    = map[string]int{
+		"spec":         12,
+		"status":       9,
+		"event":        1,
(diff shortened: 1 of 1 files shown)


operator/pkg/controllers/hubofhubs/globalhub_database.go
```diff
@@ -14,6 +17,18 @@ import (
...
+	CheckDatabaseInterval = time.Duration(10) * time.Second
+	ExpectedSchemaTables  = map[string]int{
+		"spec":         12,
+		"status":       9,
+		"event":        1,
+		"local_spec":   2,
+		"local_status": 1,
+		"history":      14,
+	}
+)
```
clyang82, line 30: it can be complex to manage the db schema. I can understand that it is cpu wasting reconciler. Can we just do the database reconciler when the controller manager is started whatever the status is? In case, the db table is deleted by accident, it can be recovered after pod starts.

yanmxa: Yes, I will add a status flag to make sure it only reconcile once.

morvencao (commented):
this looks good, but we can do better?
 - check the database condition in MGH, if it is true, then skip reconcile database, in that way, we can avoid reconcile database when operator is restarted.

yanmxa:
> this looks good, but we can do better?
> 
> * check the database condition in MGH, if it is true, then skip reconcile database, in that way, we can avoid reconcile database when operator is restarted.
> 
> Is the database reconcile expected when operator is restarted?

Yes, we expected to reconcile the database when operator is restarted.

operator/pkg/controllers/hubofhubs/globalhub_database.go
```diff
@@ -49,6 +51,14 @@ func (r *MulticlusterGlobalHubReconciler) reconcileDatabase(ctx context.Context,
 		}
 	}()
 
+	if condition.ContainConditionStatus(mgh, condition.CONDITION_TYPE_DATABASE_INIT, condition.CONDITION_STATUS_TRUE) {
```
clyang82, line 54: move it to the beginning of this reconciler method. if DatabaseReconcileCounter is greater than 0, then return nil. we do not need to get the secrets if we do not need to initiate the db.

clyang82 (changes requested):

morvencao (commented):
/lgtm

PR #465, opened 2023-06-02
fixed the bug that no history data when the event.local_polcies is empty

yanmxa:
refers: https://issues.redhat.com/browse/ACM-5772
 
 Currently, the local compliance history is synchronized with the event table. But there is a normal case that hasn't been covered:
 If a policy is created a few days ago and its compliance status isn't updated yesterday. then there is no event generated in event.local_policies on yesterday, so this policy will be dropped in history.local_compliance and that isn't the behavior we expected.


manager/pkg/cronjob/task/compliance_history.go
```diff
@@ -33,25 +34,115 @@ var (
...
+	log.Info("start running", "currentRun", job.LastRun().Format(timeFormat))
+
+	interval := dateInterval
+	if enableSimulation {
+		interval = simulationCounter
+	}
+
+	// insert or update with local_status.compliance
+	statusTotal, statusInsert, err := syncToLocalComplianceHistoryByLocalStatus(ctx, pool, batchSize, interval)
+	if err != nil {
```
clyang82, line 48: why we need call this method for non-simulation environment?

clyang82:
BTW, could you append the root cause for this ticket? it is helpful for reviewing. thanks.

manager/pkg/cronjob/task/compliance_history.go
yanmxa: I added the cause to the ticket.

yanmxa:
> BTW, could you append the root cause for this ticket? it is helpful for reviewing. thanks.

Sure, Done!

PR #470, opened 2023-06-06
use policy_id instead of id

ldpliu:


clyang82 (approved):
Good catch!

/lgtm

Thanks @ldpliu

PR #475, opened 2023-06-07
fix policy compliancy overview and some performance issue

ldpliu:
https://issues.redhat.com/browse/ACM-5843
https://issues.redhat.com/browse/ACM-5771


ldpliu:
> @ldpliu is there a reason why we are removing the `INNER JOIN` and `WHERE` clause in the template (drop down) variables? The `join` and `where` clause was to help filter the drop down list according to what was selected.

I add them back and I just want to standard the query in all of the dashboards.
Please take a look again.

clyang82 (approved):
LGTM

PR #481, opened 2023-06-09
remove unnecessary parameters from global hub api

yanmxa:
Refers: https://issues.redhat.com/browse/ACM-5848
Remove the following fields from API:
- `AggregationLevel`: set the `AggregationLevel` value always with `full`
- `MessageCompressionType`: set the value always with `gzip`, there is an issue if cloudevents haven't used it to compress the event payload. I will enhance the use of this parameter when other things are done
- `EnableLocalPolicies`: set it with always, since this feature is necessary for this delivery.


clyang82 (approved):
LGTM

PR #488, opened 2023-06-13
trigger to update manifestwork when cluster is updated

yanmxa:
Refers: https://issues.redhat.com/browse/ACM-5109


operator/pkg/controllers/addon/addon_integration_test.go
```diff
@@ -162,7 +162,7 @@ var _ = Describe("addon integration", Ordered, func() {
 				}, addon)
 			}, timeout, interval).ShouldNot(HaveOccurred())
 
-			Expect(len(addon.GetAnnotations())).Should(Equal(0))
+			// Expect(len(addon.GetAnnotations())).Should(Equal(0))
```
clyang82, line 165: Is it your desired change?

yanmxa: Yes

PR #489, opened 2023-06-13
remove managed_cluster/local_policy history tables

yanmxa:
Note:
1. remove the `history.managed_cluster` and `history.local_plicies` from `history schema`.
2. Add the trigger only update the `deleted_at` and `updated_at` when deleting row from the `status.managed_cluster` and `local_spec.policies`.
3. Make sure the deleted records(`deleted_at` IS NOT NULL) won't updated/deleted when the bundle does not contain the obj.


manager/pkg/statussyncer/syncers/local_spec_placementrule_syncer.go
```diff
@@ -19,12 +19,12 @@ import (
...
-func NewLocalSpecDBSyncer(log logr.Logger, config *corev1.ConfigMap) DBSyncer {
+// NewLocalSpecPlacementruleSyncer creates a new instance of LocalSpecDBSyncer.
+func NewLocalSpecPlacementruleSyncer(log logr.Logger, config *corev1.ConfigMap) DBSyncer {
 	dbSyncer := &LocalSpecDBSyncer{
-		log:                                     log,
-		config:                                  config,
-		createLocalPolicySpecBundleFunc:         statusbundle.NewLocalPolicySpecBundle,
+		log:    log,
+		config: config,
+		// createLocalPolicySpecBundleFunc:         statusbundle.NewLocalPolicySpecBundle,
```
clyang82, line 27: Why do we not need to create bundle?

yanmxa: Previously,`LocalPolicySpecBundle` and `LocaLPlacementRulesSpecBundle` are all in the `local_spec_db_syncer.go`.
And now `LocalPolicySpecBundle` is processed in [local_spec_policies_syncer.go](https://github.com/stolostron/multicluster-global-hub/pull/489/files/48022f37dcfeaba52bf976d894b8fcd8dff144aa#diff-ddd5c421e780d61c347a155fd2d30b1cda6ad4e19c23c4e5fdad939b1e2614c1), so I remove it from here.

clyang82 (approved):
LGTM

PR #491, opened 2023-06-15
fixed the view is not exist when simulating the local compliance

yanmxa:
When the interval is so small that the previous job has not finished running, the next job has already started.
Then a **race condition** will arise: the previous and next jobs will use the same view, which will cause the next job to fail. To avoid this, lock the counter so only one goroutine can access the it and each goroutine will get a different simulatorCounter value.

### DIFF
--- a/manager/pkg/cronjob/task/compliance_history.go
+++ b/manager/pkg/cronjob/task/compliance_history.go
@@ -3,6 +3,7 @@ package task
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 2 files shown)


morvencao (commented):
👍 
please fix the sonarcloud issue.

yanmxa:
> 👍 please fix the sonarcloud issue.

Done!

morvencao (commented):
/lgtm

PR #493, opened 2023-06-15
Updated cluster dashboard regex to ignore name label

dislbenn:
Updated the `regex` in the cluster dashboards to disable the `name` label from the `label` drop down.


bjoydeep:
The 1:10 is working very well for the `Whats changed `dashboard.
For this overview dashboard the name (and there 1:10) creates lots of problems - perhaps more than we can handle:
1. we did not allow selecting a policy in policy dashboard, so why should we add for cluster overview
1. And we are `not` limiting drop downs in the policy dashboard with time. So why here.? In fact, it is not unusual at all for people to pick and choose dimensions independent of each other and let the result deal with the intersection. And this will make the query faster (which is important!)
1. It makes it complex to forward the subset of clusters to the next level because of some known issues we have in propagating large `where cluster in () `clauses with PostgresSQL (as we are seeing with Search). 
1. also the function of this overview dashboard is to show the fleet - and not the individual cluster. For ad-hoc exploration, we have Use Case 3

So to keep matters simple, I suggest that we remove these.
@clyang82 what do you think?

clyang82:
I agree to make it simple in an initial version. We may have a better solution over time.

clyang82 (approved):
LGTM

clyang82:
FYI @hchenxa

dislbenn:
@clyang82 please help to review again.

dislbenn:
Closing this PR and will combine it with another.

PR #495, opened 2023-06-16
fixed the managed cluster deletion not work

yanmxa:
1. When we delete the managed cluster from the database, the `deletedAt` column may not be filled if the delete obj is not a pointer.
2. When the cluster is re-imported again, the `deletedAT` should be NULL.
3. Using `Find(&obj)` to get the un-deleted records, cause `db.Table(fmt.Sprintf("%s.%s", schema, tableName))` will get all the records from database.


morvencao (commented):
/lgtm

morvencao (commented):
/lgtm

PR #496, opened 2023-06-16
update dis doc

hchenxa:
refer https://issues.redhat.com/browse/ACM-4583

### DIFF
--- a/doc/disconnected_environment/README.md
+++ b/doc/disconnected_environment/README.md
@@ -1,6 +1,7 @@
 # Deploy Global Hub Operator on a Disconnected Environment 
 
 ## Prerequisites
+
 - Make sure you have an image registry, and a bastion host that has access to both the Internet and your mirror registry
 - Have OLM([Operator Lifecycle Manager](https://docs.openshift.com/container-platform/4.11/operators/understanding/olm/olm-understanding-olm.html)) installed on your cluster
 - The Advanced Cluster Management for Kubernetes has been installed on your cluster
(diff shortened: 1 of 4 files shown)


doc/disconnected_environment/README.md
```diff
@@ -69,137 +105,89 @@ spec:
   image: ${REGISTRY}/multicluster-global-hub-operator-catalog:v${VERSION}
```
yanmxa, line 105: Since you remove the line 33~55, we need to remove the `secret` from the `catalog`

doc/disconnected_environment/README.md
```diff
@@ -69,137 +105,89 @@ spec:
...
 
-$ envsubst < ./doc/disconnected-operator/catalogsource.yaml | kubectl apply -f -
+$ envsubst < ./doc/disconnected_environment/catalogsource.yaml | kubectl apply -f -
 ```
+
 OLM polls catalog sources for available packages on a regular timed interval. After OLM polls the catalog source for your mirrored catalog, you can verify that the required packages are available from on your disconnected cluster by querying the available PackageManifest resources.
+
 ```bash
-$ oc get packagemanifest -n openshift-marketplace multicluster-global-hub-operator
+$ oc get packagemanifest multicluster-global-hub-operator
```
yanmxa, line 114: Do you install the `packagemanifest` beyond `openshift-maketplace`? If not, I think it's better to add the namespace.

doc/disconnected_environment/README.md
```diff
@@ -216,27 +204,32 @@ spec:
...
+To support different image registries for different regional hubs, we can use `ManagedClusterImageRegistry` API to import the regional hub.
+
+### Config the ManagedClusterImageRegistry
 
 - Create the placement/clusterset to select the target regional hub cluster, for example:
   ```yaml
-  apiVersion: cluster.open-cluster-management.io/v1beta1
-  kind: Placement
+  apiVersion: cluster.open-cluster-management.io/v1
+  kind: ManagedCluster
```
yanmxa, line 216: Do we need to create the `ManagedCluster` when using the `ManagedClusterImageRegistry` API

hchenxa: yes, we need to create this resource and add this cluster into the clusterset so that the placement can select this managedcluster.

doc/disconnected_environment/README.md
hchenxa: we can get all packagemanifest from any namespaces.

doc/disconnected_environment/README.md
hchenxa: I put the prepare image pull secret before this steps, so we still need to set the secret here if user did not want to use the global image pull secret.

yanmxa: So do we specify a placeholder like `<global-hub-secret>` instead of a specific name?

doc/disconnected_environment/README.md
yanmxa: Is that mean we can use a fake managed cluster, Instead of joining a real managed cluster?

hchenxa: this was not the fake clusters

doc/disconnected_environment/README.md
hchenxa: updated.

doc/disconnected_environment/README.md
```diff
@@ -216,27 +204,32 @@ spec:
   dataLayer:
     type: largeScale
 ```
-The advantage of this way is that it's easy to configure. But because the **Agent image** runs in the regional hub cluster. And different clusters might have different registries, then the above method cannot reach this situation. Here is another way to achieve this.
+This was the global configuration and all of your regional hubs will use the same image registry and image pull secret.
+
+To support different image registries for different regional hubs, we can use `ManagedClusterImageRegistry` API to import the regional hub.
 
-### Config the Agent Image Registry By ManagedClusterImageRegistry
+### Config the ManagedClusterImageRegistry
```
clyang82, line 211: Do we have an existing document about how to use `ManagedClusterImageRegistry`?

hchenxa: https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.7/html-single/clusters/index#import-cluster-managedclusterimageregistry

hchenxa: hmm, may be we can add this as reference here.

hchenxa: add the reference link

PR #498, opened 2023-06-16
update the image override

hchenxa:
refer https://issues.redhat.com/browse/ACM-5985

### DIFF
--- a/operator/config/samples/operator_v1alpha3_multiclusterglobalhub.yaml
+++ b/operator/config/samples/operator_v1alpha3_multiclusterglobalhub.yaml
@@ -3,7 +3,7 @@ kind: MulticlusterGlobalHub
 metadata:
   # annotations:
     # mgh-pause: "true"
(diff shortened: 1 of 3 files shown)


operator/config/samples/operator_v1alpha3_multiclusterglobalhub.yaml
```diff
@@ -3,7 +3,7 @@ kind: MulticlusterGlobalHub
 metadata:
   # annotations:
     # mgh-pause: "true"
-    # mgh-image-repository: quay.io/<quay_io_repo>
```
clyang82, line 6: why do you remove `quay_io_repo`?

hchenxa: this should used to replace the registry server, I did not think we should support to replace the registry org.

clyang82 (approved):
LGTM

PR #504, opened 2023-06-22
fixed the local policy is not sync

yanmxa:
Refers: https://issues.redhat.com/browse/ACM-6018

- move the global hub config to the addon manifests

  I think we shouldn't treat the global hub config as the global hub resource to sync it by the global hub transport. Because whether we send the global hub resource to the transport is based on if the resource is updated. If it is not updated, the resource will not be sent to the transport.
  
  But the global hub config object must be sent to the regional hub, whether it is updated or not. So I think it should be placed in the global hub addon and deployed with the global hub agent.

- Combine interval config and global hub config into one


operator/pkg/controllers/addon/manifests/templates/agent/multicluster-global-hub-agent-cm-globalhubconfig.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+data:
+  aggregationLevel: {{ .AggregationLevel }}
+  enableLocalPolicies: "{{ .EnableLocalPolicies }}"
+metadata:
+  annotations:
+    global-hub.open-cluster-management.io/origin-ownerreference-uid: {{ .OriginOwnerReferenceUID }}
+  name: multicluster-global-hub-config
```
clyang82, line 9: multicluster-global-hub-agent-config

clyang82 (approved):
LGTM

PR #510, opened 2023-06-26
add status.leaf_hubs table to store the leaf hub information

clyang82:
https://issues.redhat.com/browse/ACM-6059

### DIFF
--- a/agent/pkg/status/controller/hubcluster/clusters_status_sync.go
+++ b/agent/pkg/status/controller/hubcluster/clusters_status_sync.go
@@ -68,28 +68,23 @@ func (c *hubClusterController) Reconcile(ctx context.Context, request ctrl.Reque
 			fmt.Errorf("reconciliation failed: %w", err)
 	}
 
-	c.syncBundle(consoleRoute)
+	c.syncBundle(ctx, consoleRoute)
 
 	reqLogger.Info("Reconciliation complete.")
 
 	return ctrl.Result{}, nil
 }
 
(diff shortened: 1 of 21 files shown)


agent/pkg/status/controller/hubcluster/clusters_status_sync.go
```diff
@@ -0,0 +1,100 @@
...
+		For(&routev1.Route{}).
+		WithEventFilter(hubClusterControllerPredicate).
+		Complete(hubClusterController); err != nil {
+		return fmt.Errorf("failed to add hub cluster controller to the manager - %w", err)
+	}
+
+	return nil
+}
+
+func (c *hubClusterController) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
```
yanmxa, line 58: Can you consider using the [generic controller](https://github.com/stolostron/multicluster-global-hub/blob/main/agent/pkg/status/controller/generic/generic_status_sync_controller.go) instead of implementing the syncBundle logic?

clyang82: I think no. the cluster info is something like the controlInfo. it is just a object. it is not a kubernetes resource so it does not have deepCopy method.

yanmxa: Actually, we need to watch the route resource and implement the update/delete method for the bundle so the generic controller can keep reconciling.  
The controlInfo doesn't reuse the generic controller because it doesn't need to watch a specific resource. So it doesn't need to reconcile.

PR #519, opened 2023-06-29
remove the not null constrait for cluster_id

yanmxa:
Refers: https://issues.redhat.com/browse/ACM-6187


clyang82:
Good catch !

/lgtm

PR #523, opened 2023-07-03
add policy ns and filter cluster_id is null data.

ldpliu:
1. Fix https://issues.redhat.com/browse/ACM-6214
Shown all policies in different namespace with the same name in policy dashboards

2. https://issues.redhat.com/browse/ACM-6213
Do not show clusters/policies which do not have cluster_id
wcp
op


hchenxa:
@ldpliu , we also need to add a known issue since we did not count the cluster when the clusterID is null, this will be difference with the view in ACM UI.

operator/pkg/controllers/hubofhubs/manifests/grafana/acm-global-offending-policies.yaml
```diff
@@ -450,7 +454,67 @@ data:
                 "properties": [
                   {
                     "id": "custom.width",
-                    "value": 259
+                    "value": 150
+                  }
+                ]
+              },
+              {
+                "matcher": {
```
clyang82, line 462: Are these your desired changes against overrides?

ldpliu: Yes, I tried to set a suitable width for each field.

ldpliu:
@clyang82  @hchenxa  Also Add a known issue in README, Please review it:
https://github.com/stolostron/multicluster-global-hub/pull/523/files#diff-a0842ea73f3a116eff7958fbf833dd2280c366c0d8985aa05f8e1403e0f39273R159

yanmxa:
@clyang82 @ldpliu @bjoydeep Do we need to add a new column `namespace` in `local_spec.policies`? Which may bring a better retrieval performance.

clyang82:
> local_spec.policies

We tested `payload -> spec -> namespace` before, the performance is fine for us. I think we do not need to add now. Thanks.

doc/README.md
```diff
@@ -154,4 +154,6 @@ For common Troubleshooting issues, proceed [here](troubleshooting.md)
 
 2. We provide ability to drill down the `Offending Policies` dashboard when you click a datapoint from the `Policy Group Compliancy Overview` dashboard. But the drill down feature is not working for the first datapoint. You can click the second datapoint or after to see the drill down feature is working. The issue is applied to the `Cluster Group Compliancy Overview` dashboard as well.
 
-3. If you detach the regional hub and then rejoin the regional hub, The data (policies/managed clusters) might not be updated in time from the rejoined regional hub. You can fix this problem by restarting the `multicluster-global-hub-manager` pod on global hub.
\ No newline at end of file
+3. If you detach the regional hub and then rejoin the regional hub, The data (policies/managed clusters) might not be updated in time from the rejoined regional hub. You can fix this problem by restarting the `multicluster-global-hub-manager` pod on global hub.
+
+4. For clusters that are never created successfully, global hub policy compliancy percentage will not count them. But these clusters will be show in RHACM policy UI.
```
clyang82, line 159: maybe we can have more details:
if the clusterclaim `id.k8s.io` does not exist in the managed cluster, then we will count this managed cluster in global hub policy compliance database. but it shows in RHACM policy console.

ldpliu: done

ldpliu: done

clyang82 (approved):
LGTM

PR #541, opened 2023-07-17
fixed the issue with BYO postgres

yanmxa:
Ref: https://issues.redhat.com/browse/ACM-6246
When bringing your own Postgres without `rootca` cert, it will throw the following bug:
```bash
2023-07-17 05:47:24 UTC ERROR   setup   failed to initialize GORM instance      {"error": "pq: SSL is not enabled on the server"}
```


clyang82:
LGTM

Issue #502, opened 2023-06-20
enhance the compliance history jobs


yanmxa:
1. Instead of `insert/update` to the compliance history table from the policy event table directly, we can `insert/update` the material view to reduce the chance of conflict, then achieve a better performance!
2. We should also filter the first row(because the prev_compliance is empty) with [this logic](https://github.com/stolostron/multicluster-global-hub/blob/9ce5dd19dc72632c3bdb640825eca5ae307f721e/manager/pkg/cronjob/task/compliance_history.go#L241-L246).

Thanks @bjoydeep for the very good advice!

Issue #536, opened 2023-07-11
Proposal for data retention/deletion with Postgres partition


yanmxa:
## Why do we need Postgres Partition?
In a nutshell, partitioning is a method used in relational databases to break down large tables into smaller partitions. This helps in executing queries/deletions on a large table faster. 

## Which type of partition do we need?
There are four main types of [partitions](https://www.postgresql.org/docs/current/ddl-partitioning.html):
- **List Partition**: Data is partitioned based on discrete values that have been specified. This method works well when you need to group discrete data, such as regions and departments, with arbitrary values.
- **Range Partition**: Data is divided into segments based on the chosen range. When you need to access time-series data, supply a date, such as a year and a month.
- **Hash Partition**: Each partition’s data is partitioned by supplying a modulus and a remainder. Each partition will contain the rows for which the modulus divided by the hash value of the partition key yields the given remainder.
- **Composite Partition**: To construct a more complex PostgreSQL partition layout, you can establish multiple partitions under a partition.

Because our scenario is like this: As time goes by, the records in the table will increase, so we want to clear out the earlier data, such as 18 months ago. Therefore, we are based on the time series to clear the data. In this case, the **Range partition** is more suitable according to the above partition definition!

## Side effects brought by range partition
- There is no option for automatically creating matching indexes on all partitions. Separate commands must be used to add indexes to each partition. This also means that there is no way to build a primary key, a unique constraint, or an exclusion constraint that spans all partitions; instead, each partition must be constrained separately.
- BEFORE ROW triggers on INSERT cannot change which partition is the final destination for a new row.
- Range partitions do not accept NULL values(before postgreSQL 11: https://www.postgresql.org/about/news/postgresql-11-released-1894)
- Before inserting data, we should make sure the partition for the record exists.

## Which tables need to be partitioned?
For those tables where data continues to grow over time
- [event.local_polcies](https://github.com/stolostron/multicluster-global-hub/blob/41123c8cc15bb45de6fdb8d5baad1d226d52a6ff/operator/pkg/controllers/hubofhubs/database/2.tables.sql#L299)
- [event.local_root_policies](https://github.com/stolostron/multicluster-global-hub/blob/41123c8cc15bb45de6fdb8d5baad1d226d52a6ff/operator/pkg/controllers/hubofhubs/database/2.tables.sql#L313C28-L313C53)
[...]

PR #543, opened 2023-07-17
add partitions to `time tables` for data retation/deletion

yanmxa:
ref: https://github.com/stolostron/multicluster-global-hub/issues/536, https://issues.redhat.com/browse/ACM-6441

- Partition tables: `event.local_policies`, `event.local_root_policies` and `history.local_compliances`
  
  - Execution the job every month and the table is partitioned monthly
  - Delete the expired(18 month ago) partition tables 
  - Create the partition table for next month
  - Record the min and max partition table of each parent table after updating them
  
- Soft deleted records: `status.managed_clusters`,  `status.leaf_hubs` and `local_spec.policies`

   - Delete the expired deleted records(18 months ago, contains the 18th)
   - Record the min `deleted_at` record kept in each table after hard deleting the records

The running logs are recorded in a table named `event.data_retention_job_log`

resolve: https://github.com/stolostron/multicluster-global-hub/issues/502


operator/pkg/controllers/hubofhubs/database/5.ptable.sql
```diff
@@ -0,0 +1,33 @@
...
+    leaf_hub_name character varying(63) NOT NULL,
+    message text,
+    reason text,
+    count integer NOT NULL DEFAULT 0,
+    source jsonb,
+    created_at timestamp without time zone DEFAULT now() NOT NULL,
+    compliance local_status.compliance_type NOT NULL,
+    CONSTRAINT local_root_policies_unique_constraint UNIQUE (event_name, count, created_at)
+) PARTITION BY RANGE (created_at);
+
```
bjoydeep, line 27: I think we should also create each partitions at this point. That is create this during installation itself. WDYT?

manager/pkg/cronjob/task/data_retention.go
```diff
@@ -0,0 +1,128 @@
...
+	end := time.Now()
+	errMessage := "none"
+	if err != nil {
+		errMessage = err.Error()
+	}
+	db := database.GetGorm()
+	result := db.Exec(`INSERT INTO event.data_retention_job_log (name, start_at, end_at, error) 
+	VALUES (?, ?, ?, ?);`, name, start, end, errMessage)
+	return result.Error
+}
```
bjoydeep, line 128: The target audience for this process are DBAs. So I think the golang code should be a wrapper around scripts or functions that need to be run. This will allow the customer DBA to inspect it and understand as well. WDYT?

operator/pkg/controllers/hubofhubs/database/5.ptable.sql
```diff
@@ -0,0 +1,33 @@
...
+CREATE TABLE IF NOT EXISTS event.data_retention_job_log (
+    name varchar(63) NOT NULL,
+    start_at timestamp NOT NULL DEFAULT now(),
+    end_at timestamp NOT NULL DEFAULT now(),
+    error TEXT
+);
```
bjoydeep, line 33: I propose we add a few more columns for clarity -

1. min-partition
2. expected min-partition
3. max-partition
4. expected max-partition
5. row-count-in-current-partition

So when we run it today - we know that current partition should be for 2023_07. We know min partition should be 2023_07 - 18 mo etc. Does that make sense?

manager/pkg/cronjob/task/data_retention.go
clyang82: Do you mean that we do not have permission to create tables so that we need to provide a script or function to allow DBAs to run? if so, I think we do not have any data once the customers install the global hub at the beginning. so we can find the real reason from the logs. Ideally, we should have our code robustly to avoid DBAs to involve in.

operator/pkg/controllers/hubofhubs/database/5.ptable.sql
yanmxa: In fact, the cronjob will run at startup to check whether there is a partition table for the current month, and if it does not exist, go back and create it. 
Yes, for the sake of reliability, we can directly create the partition table of the current month here. I added it here.

operator/pkg/controllers/hubofhubs/database/5.ptable.sql
yanmxa: Of course! So it can provide more information than just error messages

yanmxa: @bjoydeep Do we really need 2(`expected min-partition`), 4(`expected max-partition`) and 5(`row-count-in-current-partition`)? Also for the tables, like `status.managed_clusters`, with `deleted_at` columns, we should also records the min-deletion. So how about the following table?

```sql

CREATE TABLE IF NOT EXISTS event.data_retention_job_log (
    table_name varchar(63) NOT NULL,
    start_at timestamp NOT NULL DEFAULT now(),
    end_at timestamp NOT NULL DEFAULT now(),
    min_partition varchar(63), -- minimum partition after the job
    max_partition varchar(63), -- maximum partition after the job
    min_deletion  timestamp, -- the oldest deleted record keep in the table after the job
    error TEXT
);
```

operator/pkg/controllers/hubofhubs/database/5.ptable.sql
```diff
@@ -0,0 +1,80 @@
+CREATE TABLE IF NOT EXISTS event.local_policies (
```
clyang82, line 1: any concrete reason to have a separated file for partition table? if no, I suggest to keep the table schema in a single file.

manager/pkg/cronjob/sheduler.go
```diff
@@ -49,12 +49,17 @@ func AddSchedulerToManager(ctx context.Context, mgr ctrl.Manager, pool *pgxpool.
...
-	complianceJob, err := scheduler.Tag("LocalComplianceHistory").DoWithJobDetails(
-		task.SyncLocalCompliance, ctx, pool, enableSimulation)
+	complianceJob, err := scheduler.DoWithJobDetails(task.SyncLocalCompliance, ctx, pool, enableSimulation)
 	if err != nil {
 		return err
 	}
-	log.Info("set local compliance job", "scheduleAt", complianceJob.ScheduledAtTime())
+	log.Info("set SyncLocalCompliance job", "scheduleAt", complianceJob.ScheduledAtTime())
+
+	dataRetentionJob, err := scheduler.Every(1).Month(1).DoWithJobDetails(task.DataRetention, ctx, pool)
```
clyang82, line 58: the scheduler should base on the retention values. for example: if the retention is 1m, the scheduler should be running per day. Does it make sense?

manager/pkg/cronjob/task/data_retention.go
```diff
@@ -0,0 +1,180 @@
...
+var (
+	// The main tasks of this job are:
+	// 		1. create partition tables for days in the future, the partition table for the next month is created
+	//    2. delete partition tables that are no longer needed, the partition table for the previous 18 month is deleted
+	//    3. completely delete the soft deleted records from database after retainedMonths
+	retentionTaskName = "data-retention"
+
+	// after the record is marked as deleted, retentionMonth is used to indicate how long it will be retained
+	// before it is completely deleted from database
+	retentionMonth  = 18
```
clyang82, line 26: the retention should be from the value which the customer is configured in MCGH CR.

operator/pkg/controllers/hubofhubs/database/5.ptable.sql
yanmxa: I place these two types of tables in different files to distinguish they are not exactly the same type. Of course! we can keep them in one file.

manager/pkg/cronjob/sheduler.go
yanmxa: @clyang82 Not necessarily! 
It makes sense. I agree with you that the job should be running more than once for the duration of the retention. But running every day seems too frequent. How about every week?

manager/pkg/cronjob/task/data_retention.go
yanmxa: Yes. Just to declare here temporarily.

manager/pkg/cronjob/task/data_retention.go
```diff
@@ -0,0 +1,180 @@
...
+)
+
+func DataRetention(ctx context.Context, pool *pgxpool.Pool, job gocron.Job) {
+	currentTime := time.Now()
+	log = ctrl.Log.WithName(retentionTaskName)
+
+	creationTime := currentTime.AddDate(0, 1, 0)
+	deletionTime := currentTime.AddDate(0, -retentionMonth, 0)
+	for _, tableName := range partitionTables {
+		err := updatePartitionTables(tableName, creationTime, deletionTime)
```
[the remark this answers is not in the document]
yanmxa: @clyang82 I have traced the job status in the following line(line 52) with `traceDataRetentionLog()` function. Do you mean split the PR?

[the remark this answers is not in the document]
clyang82: Just logged this story for emitting metrics https://issues.redhat.com/browse/ACM-6708

clyang82:
I am OK for letting this PR go.
/lgtm

PR #556, opened 2023-07-25
Enhance e2e tests

clyang82:
1. reduce the wait time can save 3 min for e2e.
2. refactor the option structure to `globalhub` --> `managedhubs` --> `managedclusters`
3. remove microshift since the `microshift-aio` image is deprecated.
4. no authentication required for non-k8s api since no microshift here. introduce `mgh-cluster-api-url` annotation to set `cluster-api-url` as empty bypass the authentication.
5. use `kustomize` api to deploy the multicluster global hub operator instead of executing command in golang code.
6. deploy the service-ca to continue using the serving certs.
7. remove OLM dependency in e2e


clyang82:
Before:
- Prepare top hub :82 seconds
- Enable OLM :57 seconds
- initHub :0 seconds
- Prepare leaf hub cluster :661 seconds
- Init HoH OCM :1 seconds
- check connection :253 seconds
- wait kafka :0 seconds
- Ran 32 of 39 Specs in 560.285 seconds
- Ran 7 of 39 Specs in 171.510 seconds

clyang82:
continue to investigate why kafka spend 10m to ready

clyang82:
AFTER:
Prepare top hub :78 seconds
Prepare leaf hub cluster :375 seconds
check connection :109 seconds
Ran 31 of 38 Specs in 516.191 seconds
Ran 7 of 38 Specs in 162.815 seconds

test/pkg/e2e/suite_test.go
```diff
@@ -157,32 +143,19 @@ func deployGlobalHub() {
...
+	Expect(kustomize.Apply(testClients, testOptions,
+		kustomize.Options{KustomizationPath: fmt.Sprintf("%s/operator/config/default", rootDir)})).NotTo(HaveOccurred())
 
 	By("Deploying operand")
 	mcgh := &operatorv1alpha3.MulticlusterGlobalHub{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "multiclusterglobalhub",
 			Namespace: "open-cluster-management",
+			Annotations: map[string]string{
+				constants.AnnotationMGHClusterAPIUrl: "test",
```
yanmxa, line 157: It seems that the `AnnotationMGHClusterAPIUrl` is only for the e2e test. 
Why do we need this annotation with the `test` value?

yanmxa: Also, I saw that the argument `--cluster-api-url` has been passed to the manager. But it seems no manager updates were committed. So I can't figure out how it works.

clyang82: refer to: https://github.com/stolostron/multicluster-global-hub/blob/508ec1ea254706bb229adda0a68f62895cb5533d/manager/pkg/nonk8sapi/non_k8s_api.go#L99

yanmxa: if the `--cluster-api-url` is only for skip authentication. How about renaming it to a readable argument, like `skip-auth`?

clyang82: Good suggestion. I have renamed the annotation to `mgh-skip-auth` to skip auth for non-k8s api.

PR #557, opened 2023-07-25
forbidd the cluster policy events from managed hub

yanmxa:
Ref: https://issues.redhat.com/browse/ACM-6593

The following cluster policy events will reported after I create an informed policy and then enforce it:

image

However, we can find that the red boxes in the figure are not helpful for analyzing the reasons for the policy state, so it is recommended to remove them.


yanmxa:
add the jira to trace this updates

PR #559, opened 2023-07-27
add postgres data retention to the global hub operand

yanmxa:
Ref: https://issues.redhat.com/browse/ACM-6256


operator/apis/v1alpha3/multiclusterglobalhub_types.go
```diff
@@ -99,6 +99,15 @@ type DataLayerConfig struct {
...
 	Kafka *KafkaConfig `json:"kafka,omitempty"`
+	// +optional
+	Postgres *PostgresConfig `json:"postgres,omitempty"`
+}
+
+// PostgresConfig defines the desired state of postgres
+type PostgresConfig struct {
+	// Expiration is the time duration before the data is deleted from the database. default value is 18 months
+	// +kubebuilder:default:=18
+	Expiration int `json:"expiration,omitempty"`
```
yanmxa, line 110: `1d`, `1m` and `1y`

operator/config/crd/bases/operator.open-cluster-management.io_multiclusterglobalhubs.yaml
```diff
@@ -60,6 +60,19 @@ spec:
...
+                        description: PostgresConfig defines the desired state of postgres
+                        properties:
+                          retention:
+                            default: 18m
+                            description: Retention is a a duration string. Which defines
+                              how long to keep the data in the database. min value
+                              is 1 month. A duration string is a possibly signed sequence
+                              of decimal numbers, each with optional fraction and
+                              a unit suffix, such as "2d4m". Valid time units are
+                              "h", "d", "m", "y".
```
clyang82, line 73: if the minimum value  is `1m`, then we do not need to support `h` and `d` I think. Do we need to support `2m1d`? maybe not.

yanmxa: I thought there might be some special scenes, such as `1m14d`, However, this does not seem to be common.

clyang82: maybe we change `min value is 1 month` to `recommended minimum value is 1 month`

pkg/testdata/crds/0000_03_operator.open-cluster-management.io_multiclusterglobalhubs.yaml
```diff
@@ -67,40 +59,35 @@ spec:
...
+                      postgres:
+                        description: PostgresConfig defines the desired state of postgres
+                        properties:
+                          retention:
+                            default: 18m
+                            description: Retention is a a duration string. Which defines
+                              how long to keep the data in the database. min value
+                              is 1 month. A duration string is a possibly signed sequence
+                              of decimal numbers, each with optional fraction and
+                              a unit suffix, such as "2d4m". Valid time units are
```
clyang82, line 71: you need to regenerate the crds.

operator/pkg/condition/condition.go
```diff
@@ -52,6 +52,12 @@ const (
 	CONDITION_MESSAGE_DATABASE_INIT = "Database has been initialized"
 )
 
+// NOTE: the status of Data Retention can be True or False
+const (
+	CONDITION_TYPE_RETENTION_PARSED   = "DataRetentionParsed"
+	CONDITION_REASON_RETENTION_PARSED = "DataRetentionParsed"
```
clyang82, line 58: what is real message if the user set an unsupported value? e.g.: 1m10d

pkg/testdata/crds/0000_03_operator.open-cluster-management.io_multiclusterglobalhubs.yaml
yanmxa: Done.

operator/pkg/condition/condition.go
yanmxa: It will add the error message to the `DataRetentionParsed` condition:
`time: unknown unit 'd' in duration '1m10d'`

PR #560, opened 2023-07-27
add investigation help dashboard

ldpliu:


ldpliu:
/cc @clyang82 @bjoydeep 
Please review the new dashboard for usecase3.
This is the video with this pr: https://drive.google.com/file/d/1lolVbcVjYkDznlUL7vT7TN25mLP0XB2M/view?usp=sharing

clyang82:
Awesome job! Thanks @ldpliu some questions and concerns:
1. I do not think the dashboard name `Investigation Help` is a final name. Maybe we call `Global Hub - Offending Policy`
2. Since we have `Offending clusters (Group by message)`, the `Offending Clusters` table can be redundant.
3. Maybe remove the Unknown Clusters table if there are no any messages available. 
4. Can we make the `Cluster Count` clickable?

ldpliu:
> Awesome job! Thanks @ldpliu some questions and concerns:
> 
>     1. I do not think the dashboard name `Investigation Help` is a final name. Maybe we call `Global Hub - Offending Policy`
> 
>     2. Since we have `Offending clusters (Group by message)`, the `Offending Clusters` table can be redundant.
> 
>     3. Maybe remove the Unknown Clusters table if there are no any messages available.
> 
>     4. Can we make the `Cluster Count` clickable?

Thanks for your comments.
For 1: As we already have a dashboard named `Global Hub - Offending Policies`, So May be we could call it `Global Hub - Offending One Policy` ? or @bjoydeep Could you help to suggest a better name?
For 2, Currently, user could click the `policy` and `cluster` and drill down to the `what's change cluster/policy` page to list all the events about this policy in this cluster. I am not sure if the behavior is useful to Customer. @bjoydeep any suggestion?
For 3, Yes, it's only used to list the clusters name. 
For 4: I will make the clustercount clickable, and open the ACM policy detail in leafhub.

bjoydeep:
This is very good work!

1. Can we use `Ad hoc Policy investigation`
2. I actually looked up a message from `Offending clusters (Group by message)` and then looked up `Offending Clusters` view to find out the clusters that had that message. So I already used both the views (even before looking at the comment)
3. Yes, lets remove Unknown if there are no messages available.

Let's get this going. After checking this in, we can start to review this with others.

clyang82:
I am OK to let this PR go. But I quick comment is can we add a underline for the link?

clyang82 (approved):
LGTM

PR #561, opened 2023-07-27
remove postgres/kafka installation from shell

yanmxa:
Ref: https://issues.redhat.com/browse/ACM-4877


yanmxa:
/hold 
The `postgres/kafka` in e2e may be enahnced in this PR: https://github.com/stolostron/multicluster-global-hub/pull/556

clyang82:
we will keep the postgres and kafka as it is. it belongs to middleware. not part of mgch.

PR #563, opened 2023-08-01
reset the bundle version when the agent start without the version info

yanmxa:
ref: https://issues.redhat.com/browse/ACM-6258


clyang82:
Please remove the known issue about this issue in document since we have fixed it in 1.0.0

PR #567, opened 2023-08-03
add cronjob status to metrics server

yanmxa:
ref: https://issues.redhat.com/browse/ACM-6708


yanmxa:
The e2e test failed caused by the [KinD AMI](https://github.com/stolostron/cicd-docs/blob/main/prow/kind_tests.md#the-acm-kind-ami).

ref: https://redhat-internal.slack.com/archives/CSZLMKPS5/p1691086669266829

yanmxa:
@clyang82 /cc

operator/pkg/controllers/hubofhubs/globalhub_config.go
```diff
@@ -17,17 +17,40 @@ import (
...
 ) error {
 	log := r.Log.WithName("config")
+	log.Info("set operand images; monitor the global hub namespace; set global hub agent config")
 	// set image overrides
-	log.Info("set operand images; reconcile global hub configmap")
 	if err := config.SetImageOverrides(mgh); err != nil {
 		return err
 	}
 
+	// add label openshift.io/cluster-monitoring: "true" to the ns, so that the prometheus can detect the ServiceMonitor.
```
clyang82, line 33: I am looking for the best practices here. right now, we have 2 approaches:
1. add label to namespace to allow prometheus scrapping metrics based on the definition of servicemonitor. it seems it applies to all `openshift-*` namespace.
2. another one is to drop the servicemonitor into openshift-monitoring namespace. that is what ACM does today.

clyang82: let us directly drop our servicemonitor into openshift-monitoring namespace. thanks.

yanmxa: Thanks @clyang82! It's better to move the global hub `servicemonitor` to `openshift-monitoring`, rather than adding the monitor label to the global hub namespace.

clyang82 (approved):
Great job! Thanks

PR #569, opened 2023-08-07
fixed the nil pointer bug when the operand field is missing

yanmxa:
ref: https://issues.redhat.com/browse/ACM-6780


clyang82:
@yanmxa Do we have the default value for the retention? in other words, is this default value set - https://github.com/stolostron/multicluster-global-hub/blob/f8719071dbc6e050a225a754657d2f9ed72bda9c/operator/apis/v1alpha3/multiclusterglobalhub_types.go#L112C1-L112C32

yanmxa:
> 

Yes, we have the default value for `retention`. But haven't the default value for postgreConfig pointer: https://github.com/stolostron/multicluster-global-hub/blob/f8719071dbc6e050a225a754657d2f9ed72bda9c/operator/apis/v1alpha3/multiclusterglobalhub_types.go#L103

yanmxa:
I close the PR, since after simplifying the Global Hub Operand in this [PR](https://github.com/stolostron/multicluster-global-hub/pull/576). The error will disappear~

PR #574, opened 2023-08-10
Create readonly user

clyang82:
create `guest` as readonly user for the postgres. it is used for query data from grafana.


yanmxa:
I remember we discussed that the `gust` user should be created by the operator. The customer doesn't need to provide the guest URI.

clyang82:
we do not have a place to store the password for `guest` unless we hardcode a password. Let us collect the customer's feedback to enhance it later.

PR #577, opened 2023-08-14
Refactor the `bundleversion` to support resetting the version of conflation unit

yanmxa:
Cause the agent syncs the bundle periodically, it might skip the minimum version(`0.1`) for the bundle when sending messages to transport. 
The min version is also [a sign of resetting the bundle version](https://github.com/stolostron/multicluster-global-hub/pull/563) for the manager. The skipping version will result in bundles being filtered out based on the version, and eventually, the latest bundle cannot be updated in the database.

The solution is to keep the bundle version from the agent with a minimum value(`0.1`) when it first time sends the bundle to transport.

At last, with the above mechanism, I think we can consider deprecating the `incarnation` from the bundle. That means the agent doesn't need the `incarnation-config` configmap anymore. 

### Goals

- Refactor the bundle version
  - Update the version schema, which allows us not to add another message type to reset the bundle version for manager or add a flag to stop the increment of the bundle version for the agent.
  ``` go
  // BundleVersion holds the information necessary for the consumers of status bundles to compare versions correctly.
  type BundleVersion struct {
	  // this Value is incremented every time the bundle is sent to the hub
	  Generation uint64 `json:"Generation"`
	  // this Value is incremented every time the bundle is updated
	  Value uint64 `json:"Value"`
  }
  ```
  - Remove the incarnation configmap from the managed hub
  - The manager reset the last processed bundle version when getting the Generation with `0`
      
- Refactor the Policy Compliance Bundles to Database

   - Replace the hard-to-read SQL splicing with an ORM to interact with the database(make it easier to troubleshoot), refer: https://github.com/stolostron/multicluster-global-hub/pull/586
   
   - There are different bundles to update the same table, cut off the redundant bundle to reduce the workload


yanmxa:
Current 
1. agent 0.27              -> manager 0.27
2. Agent 0.1                -> manager 0.1 -> 0.0 ===
3. Agent 0.2               -> manger 0.2 > 0.1 ===


Before 
1. agent 0.27              -> manager 0.27
2. agent 1.0                -> manager 1.0 > 0.27 ========
3. agent 1.1                 -> manager 1.1 > 1.0 ========

clyang82 (approved):

PR #581, opened 2023-08-15
Consolidate the database worker pool

yanmxa:
This focuses on the following points:

1. Replace the `time.after` with a `ticker` to fix the memory leak problem. For details, refer to the [blog](https://www.arangodb.com/2020/09/a-story-of-a-memory-leak-in-go-how-to-properly-use-time-after).
2. The global hub manager might crash when the [conflation dispatcher](https://github.com/stolostron/multicluster-global-hub/blob/5ef47947e5613ed942c305b6b7f07380c8df754f/manager/pkg/statussyncer/dispatcher/conflation_dispatcher.go#L51) gets the `DBworker` from the `worker pool` with 10 seconds timeout. A suitable way is to add a retry mechanism to avoid the failure caused by race conditions during initialization.
3. This also fixes the problem that the unit tests for global hub manager syncers are unstable.

Ref: https://github.com/stolostron/multicluster-global-hub/pull/577


clyang82 (approved):
Good catch!

PR #583, opened 2023-08-16
Modify the global hub sample with API v1alpha4

yanmxa:


hchenxa:
@yanmxa please update https://github.com/stolostron/multicluster-global-hub/blob/main/operator/config/samples/kustomization.yaml as well.

morvencao (commented):
/lgtm

PR #630, opened 2023-09-13
Replace the duration with month(int) for the data retention

yanmxa:
ref: https://issues.redhat.com/browse/ACM-7168

### DIFF
--- a/pkg/utils/time_test.go
+++ b/pkg/utils/time_test.go
@@ -2,6 +2,7 @@ package utils
 
 import (
 	"errors"
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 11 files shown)


pkg/utils/time_test.go
```diff
@@ -67,3 +68,35 @@ func TestParseDuration(t *testing.T) {
 		})
 	}
 }
+
+func TestParseRetention(t *testing.T) {
+	s := "-1y4m"
```
clyang82, line 73: should be parsed error

yanmxa: Done.

PR #634, opened 2023-09-14
Disable global resource support

clyang82:
ref: https://issues.redhat.com/browse/ACM-6564


test/pkg/testpostgres/testdatabase.go
```diff
@@ -52,5 +52,27 @@ func InitDatabase(uri string) error {
 		}
 		fmt.Printf("script %s executed successfully.\n", file.Name())
 	}
+
+	sqlDir = filepath.Join(dirname, "operator", "pkg", "controllers", "hubofhubs", "database.old")
```
yanmxa, line 56: Can we need to remove the old sqls from here?

I'm worried about the old SQL will override the updated sql, which may caused the update skips from the UT

yanmxa: A: the old sql if for global resource

yanmxa:
Also the e2e does not cover global resource cases.

PR #640, opened 2023-09-18
Fixed the bug that multiclusterglobalhub status is not updated

yanmxa:
Ref: https://issues.redhat.com/browse/ACM-7428


morvencao (commented):
/lgtm

PR #644, opened 2023-09-19
Add annotation to trigger the cronjobs

yanmxa:


manager/cmd/manager/main.go
```diff
@@ -139,7 +141,9 @@ func parseFlags() *managerconfig.ManagerConfig {
 	pflag.IntVar(&managerConfig.DatabaseConfig.DataRetention, "data-retention", 18,
 		"data retention indicates how many months the expired data will kept in the database")
 	pflag.BoolVar(&managerConfig.EnableGlobalResource, "enable-global-resource", false,
-		"Enable the global resource feature.")
+		"enable the global resource feature.")
+	pflag.StringVar(&managerConfig.LaunchJobImmediately, "launch-job-immediately", "",
+		"the specified jobs(concatenate multiple jobs with ',') runs when the container starts")
```
clyang82, line 146: Can we not expose in command? expose in command means user can use it. if we expose an env, nobody knows that. I mean that we can add an env in the manager deployment for this purpose. And then use `os.Getenv` to get the value.

clyang82 (changes requested):
BTW, I just found that the `sheduler.go` file name should be `scheduler.go`.

manager/cmd/manager/main.go
yanmxa: Done.

manager/cmd/manager/main.go
```diff
@@ -40,12 +40,13 @@ import (
...
-	webhookCertDir             = "/webhook-certs"
-	kafkaTransportType         = "kafka"
-	leaderElectionLockID       = "multicluster-global-hub-manager-lock"
+	metricsHost                   = "0.0.0.0"
+	metricsPort             int32 = 8384
+	webhookPort                   = 9443
+	webhookCertDir                = "/webhook-certs"
+	kafkaTransportType            = "kafka"
+	leaderElectionLockID          = "multicluster-global-hub-manager-lock"
+	launchJobImmediatelyEnv       = "LAUNCH_JOB_IMMEDIATELY"
```
clyang82, line 49: what is the value of `LAUNCH_JOB_IMMEDIATELY`? in literal, the value should be true or false.

yanmxa: Since we have two jobs, the value is the cronjob name. We can specify the name to launch the job when the container starts.

clyang82: maybe `launch_job_names`?

yanmxa: Done.

clyang82 (commented):
/lgtm

PR #646, opened 2023-09-19
Wait the kafka ready to apply global hub addon

yanmxa:
Ref: https://issues.redhat.com/browse/ACM-7570


yanmxa:
The bug happens after the initialization. So the fix is not correct.

PR #655, opened 2023-09-27
fix too much log issue

ldpliu:
https://issues.redhat.com/browse/ACM-5096

### DIFF
--- a/operator/pkg/controllers/addon/manifests/templates/agent/multicluster-global-hub-agent-deployment.yaml
+++ b/operator/pkg/controllers/addon/manifests/templates/agent/multicluster-global-hub-agent-deployment.yaml
@@ -28,7 +28,7 @@ spec:
               memory: 1000Mi
           imagePullPolicy: {{ .ImagePullPolicy }}
           args:
-            - '--zap-devel=true'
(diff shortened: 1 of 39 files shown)


ldpliu:
> I think we should control the log level by parameter. Avoid reducing the log by changing all the levels of the log info.
> 
> What do you think of it?

Yes, you are right.
Changed. https://github.com/stolostron/multicluster-global-hub/pull/655/files#diff-f7893cd19dbd21bb6025ab8336c0fa1799f30e7cfcd17fd0cbcb73c7c0d1e3b9R253
Please review again.

operator/pkg/controllers/addon/manifests/templates/agent/multicluster-global-hub-agent-deployment.yaml
```diff
@@ -22,7 +22,8 @@ spec:
           image: {{ .HoHAgentImage }}
           imagePullPolicy: {{ .ImagePullPolicy }}
           args:
-            - '--zap-devel=true'
+            - --zap-devel=true
```
morvencao, line 25: we may consider to disable zap dev mode to reduce the logs.

clyang82 (approved):

PR #665, opened 2023-10-08
Watch the image pull secret and propagate it to components

yanmxa:
Ref: https://issues.redhat.com/browse/ACM-4732

### DIFF
--- a/operator/pkg/controllers/hubofhubs/globalhub_config.go
+++ b/operator/pkg/controllers/hubofhubs/globalhub_config.go
@@ -18,6 +18,8 @@ func (r *MulticlusterGlobalHubReconciler) reconcileSystemConfig(ctx context.Cont
 		return err
 	}
 
+	config.SetImagePullSecretName(mgh)
+
 	// set statistic log interval
 	if err := config.SetStatisticLogInterval(mgh); err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


yanmxa:
override "SonarCloud Code Analysis"

operator/pkg/controllers/hubofhubs/globalhub_config.go
```diff
@@ -18,6 +18,8 @@ func (r *MulticlusterGlobalHubReconciler) reconcileSystemConfig(ctx context.Cont
 		return err
 	}
 
+	config.SetImagePullSecretName(mgh)
```
morvencao, line 21: can we do this in predicate function to avoid the race condition and missing call?

yanmxa: @morvencao Thanks for your review! The config reconciling will always be in front of the manager and Grafana reconciles (which will use the image pull secret name). So I think the race condition may not exist in such case. What do you think of it?

image

morvencao: it makes sense.

morvencao:
leave this to @clyang82 and @ldpliu for future review.

PR #668, opened 2023-10-09
Update global-hub-user password

clyang82:
move the create role and alter role into postgres start script to be executed when postgres pod starts. in the past, it needed the admin user have create role permission which may not be required.

### DIFF
--- a/operator/pkg/controllers/hubofhubs/manifests/postgres/init-configmap.yaml
+++ b/operator/pkg/controllers/hubofhubs/manifests/postgres/init-configmap.yaml
@@ -7,3 +7,5 @@ data:
   postgresql-start.sh: |
 
     psql -U postgres -tc "SELECT 1 FROM pg_database WHERE datname = 'hoh'" | grep -q 1 || psql -U postgres -c "CREATE DATABASE hoh"
+    psql -U postgres -tc "SELECT 1 FROM pg_catalog.pg_roles WHERE rolname = '{{.PostgresReadonlyUsername}}'" | grep -q 1 || psql -U postgres -c "CREATE ROLE \"{{.PostgresReadonlyUsername}}\" LOGIN PASSWORD '{{.PostgresReadonlyUserPassword}}'"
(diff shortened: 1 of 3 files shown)


operator/pkg/controllers/hubofhubs/manifests/postgres/init-configmap.yaml
```diff
@@ -7,3 +7,5 @@ data:
   postgresql-start.sh: |
 
     psql -U postgres -tc "SELECT 1 FROM pg_database WHERE datname = 'hoh'" | grep -q 1 || psql -U postgres -c "CREATE DATABASE hoh"
+    psql -U postgres -tc "SELECT 1 FROM pg_catalog.pg_roles WHERE rolname = '{{.PostgresReadonlyUsername}}'" | grep -q 1 || psql -U postgres -c "CREATE ROLE \"{{.PostgresReadonlyUsername}}\" LOGIN PASSWORD '{{.PostgresReadonlyUserPassword}}'"
+    psql -U postgres -c "ALTER ROLE \"{{.PostgresReadonlyUsername}}\" WITH PASSWORD '{{.PostgresReadonlyUserPassword}}'"
```
yanmxa, line 11: So we will only update the password of `PostgresReadonlyUser`. And the global-hub-user password will not update?

clyang82: we have 2 users. one is `postgres` which is a superuser. the other is `global-hub-readonly-user`. the `postgres` user password can be updated via environment - https://github.com/stolostron/multicluster-global-hub/blob/ba8928b2f2e286d4b71fee5a64b0d8e1d3ddc215/operator/pkg/controllers/hubofhubs/manifests/postgres/statefulset.yaml#L22. `global-hub-readonly-user` user will be updated here.

yanmxa: Thanks!

PR #682, opened 2023-10-17
Update 5.privileges.sql for global resource case

clyang82:
need double quotes for the user whose name includes `-`.


clyang82:
/hold

will have more fixes to make the global resource path works

PR #684, opened 2023-10-17
fix reconcile too long and some error in log

ldpliu:
https://issues.redhat.com/browse/ACM-7913
1. When Kafka or Postgre is starting, we should not return errors in reconcile. We just need to wait it started
2. We should not return `mgh is not exist` errors before `mgh` is created. just need to wait it created.

### DIFF
--- a/operator/pkg/condition/condition.go
+++ b/operator/pkg/condition/condition.go
@@ -74,6 +74,14 @@ const (
 	CONDITION_MESSAGE_LEAFHUB_DEPLOY_FAILED = "Leaf Hub Deployed FAILED"
 )
 
+const (
+	CONDITION_TYPE_GLOBALHUB_READY    = "Ready"
+	CONDITION_REASON_GLOBALHUB_READY  = "MulticlusterGlobalHubReady"
+	CONDITION_MESSAGE_GLOBALHUB_READY = "Multicluster Global Hub is ready"
+	CONDITION_REASON_GLOBALHUB_FAILED = "MulticlusterGlobalHubFailed"
+	CONDITION_REASON_GLOBALHUB_PAUSED = "MulticlusterGlobalHubPaused"
+)
+
 // SetConditionFunc is function type that receives the concrete condition method
 type SetConditionFunc func(ctx context.Context, c client.Client,
 	mgh *globalhubv1alpha4.MulticlusterGlobalHub,
--- a/operator/pkg/controllers/addon/addon_controller.go
+++ b/operator/pkg/controllers/addon/addon_controller.go
@@ -2,6 +2,7 @@ package addon
 
 import (
 	"context"
+	"time"
 
 	"github.com/go-logr/logr"
 	operatorsv1 "github.com/operator-framework/api/pkg/operators/v1"
@@ -18,6 +19,8 @@ import (
 	addonv1alpha1client "open-cluster-management.io/api/client/addon/clientset/versioned"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/stolostron/multicluster-global-hub/operator/pkg/utils"
 
 	globalhubv1alpha4 "github.com/stolostron/multicluster-global-hub/operator/apis/v1alpha4"
 	operatorconstants "github.com/stolostron/multicluster-global-hub/operator/pkg/constants"
@@ -111,7 +114,10 @@ func (a *HoHAddonController) Start(ctx context.Context) error {
 		ControllerConfig:     a.ControllerConfig,
 		LogLevel:             a.LogLevel,
 	}
-
+	_, err = utils.WaitGlobalHubReady(ctx, a.client, 5*time.Second)
+	if err != nil {
+		return err
+	}
 	agentAddon, err := addonfactory.NewAgentAddonFactory(
 		operatorconstants.GHManagedClusterAddonName, FS, "manifests").
 		WithAgentHostedModeEnabledOption().
--- a/operator/pkg/controllers/hubofhubs/globalhub_controller.go
+++ b/operator/pkg/controllers/hubofhubs/globalhub_controller.go
@@ -33,9 +33,12 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
 	"open-cluster-management.io/addon-framework/pkg/addonmanager"
 	addonv1alpha1 "open-cluster-management.io/api/addon/v1alpha1"
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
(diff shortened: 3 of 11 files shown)


clyang82:
Could you append the root cause why the reconciler takes too long in the description? Thanks.

clyang82:
It may be related with `DefaultControllerRateLimiter` https://github.com/kubernetes-sigs/controller-runtime/blob/15d79283523517f52ee30b28d2ae4c1ea3f113da/pkg/controller/controller.go#L146. 

NewItemExponentialFailureRateLimiter(5*time.Millisecond, 1000*time.Second), the maxDelay is 1000 second. that can be a root cause. I would like to have a try with customized option in controller manager https://github.com/stolostron/multicluster-global-hub/blob/7da17d464be9b1db46702f1790200245c7ae99ab/operator/pkg/controllers/hubofhubs/globalhub_controller.go#L482 with the following code:

```
WithOptions(controller.Options{
			RateLimiter: workqueue.NewItemFastSlowRateLimiter(5*time.Millisecond, time.Second*5, 5),
		}).
```

ldpliu:
> It may be related with `DefaultControllerRateLimiter` https://github.com/kubernetes-sigs/controller-runtime/blob/15d79283523517f52ee30b28d2ae4c1ea3f113da/pkg/controller/controller.go#L146.
> 
> NewItemExponentialFailureRateLimiter(5_time.Millisecond, 1000_time.Second), the maxDelay is 1000 second. that can be a root cause. I would like to have a try with customized option in controller manager
> 
> https://github.com/stolostron/multicluster-global-hub/blob/7da17d464be9b1db46702f1790200245c7ae99ab/operator/pkg/controllers/hubofhubs/globalhub_controller.go#L482
> with the following code:
> 
> ```
> WithOptions(controller.Options{
> 			RateLimiter: workqueue.NewItemFastSlowRateLimiter(5*time.Millisecond, time.Second*5, 5),
> 		}).
> ```

Yes, I append some details in description.
But I think the root cause is we returned too much errors when we wait `kafka or postgre` starting. We just need to wait they started instead of always return errors .

operator/pkg/utils/utils.go
```diff
@@ -252,3 +257,37 @@ func CopyMap(newMap, originalMap map[string]string) {
 		newMap[key] = value
 	}
 }
+
+func WaitGlobalHubReady(ctx context.Context,
+	client client.Client,
+	interval time.Duration) (*globalhubv1alpha4.MulticlusterGlobalHub, error) {
+	mghInstance := &globalhubv1alpha4.MulticlusterGlobalHub{}
+
+	if err := wait.PollInfinite(interval, func() (bool, error) {
```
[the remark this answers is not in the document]
ldpliu: > Should we anticipate the check running indefinitely if certain pods are not ready? Perhaps adding a timeout would be a better approach.

Done

operator/pkg/utils/utils.go
```diff
@@ -252,3 +257,37 @@ func CopyMap(newMap, originalMap map[string]string) {
...
+		mghInstance = &mghList.Items[0]
+		if !mghList.Items[0].DeletionTimestamp.IsZero() {
+			klog.Info("MulticlusterGlobalHub is deleting")
+			return false, nil
+		}
+
+		if meta.IsStatusConditionTrue(mghInstance.Status.Conditions, condition.CONDITION_TYPE_GLOBALHUB_READY) {
+			return true, nil
+		}
+		klog.Info("Wait MulticlusterGlobalHub is ready")
```
clyang82, line 287: maybe just print the logs once at the top of this method to tell user we are waiting for mcgh ready. It is too noisy to print every 10 seconds.

operator/pkg/controllers/hubofhubs/globalhub_kafka.go
```diff
@@ -135,6 +135,9 @@ func (r *MulticlusterGlobalHubReconciler) WaitForKafkaClusterReady(ctx context.C
 		Namespace: config.GetDefaultNamespace(),
 	}, kafkaCluster)
 	if err != nil {
+		r.Log.Error(err, "Failed to get Kafka cluster",
```
clyang82, line 138: Can we add log level here?

operator/pkg/controllers/hubofhubs/globalhub_controller.go
```diff
@@ -131,31 +133,40 @@ type MiddlewareConfig struct {
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.0/pkg/reconcile
 func (r *MulticlusterGlobalHubReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	if len(req.Namespace) == 0 || len(req.Name) == 0 {
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+	}
```
clyang82, line 138: what is the case that the req.Name or Namespace can be empty?

operator/pkg/condition/condition.go
```diff
@@ -74,6 +74,14 @@ const (
 	CONDITION_MESSAGE_LEAFHUB_DEPLOY_FAILED = "Leaf Hub Deployed FAILED"
 )
 
+const (
+	CONDITION_TYPE_GLOBALHUB_READY    = "Ready"
+	CONDITION_REASON_GLOBALHUB_READY  = "MulticlusterGlobalHubReady"
+	CONDITION_MESSAGE_GLOBALHUB_READY = "Multicluster Global Hub is ready"
+	CONDITION_REASON_GLOBALHUB_FAILED = "MulticlusterGlobalHubFailed"
+	CONDITION_REASON_GLOBALHUB_PAUSED = "MulticlusterGlobalHubPaused"
+)
```
clyang82, line 83: I like the idea to add `Ready` condition to indicate the global hub is ready for process. Thank you

operator/pkg/controllers/addon/addon_controller.go
```diff
@@ -111,7 +114,10 @@ func (a *HoHAddonController) Start(ctx context.Context) error {
 		ControllerConfig:     a.ControllerConfig,
 		LogLevel:             a.LogLevel,
 	}
-
+	_, err = utils.WaitGlobalHubReady(ctx, a.client, 10*time.Second)
```
clyang82, line 117: will it too long? can we check it in 5 seconds like what we did in the past?

operator/pkg/controllers/hubofhubs/globalhub_controller.go
```diff
@@ -131,31 +133,40 @@ type MiddlewareConfig struct {
...
 			return ctrl.Result{}, nil
 		}
 		// Error reading the object - requeue the request.
 		r.Log.Error(err, "Failed to get MulticlusterGlobalHub")
 		return ctrl.Result{}, err
 	}
-
 	if config.IsPaused(mgh) {
 		r.Log.Info("mgh reconciliation is paused, nothing more to do")
+		conditionErr := AddFailedCondition(ctx, r.Client, mgh, "mgh reconciliation is paused")
```
clyang82, line 156: we do not need to handle the pause case. it is only for developer. And `pause` does not mean the condition is failed.

operator/pkg/controllers/hubofhubs/globalhub_controller.go
```diff
@@ -167,12 +178,41 @@ func (r *MulticlusterGlobalHubReconciler) Reconcile(ctx context.Context, req ctr
...
+		return ctrl.Result{}, err
+	}
+	//Watch middleware avaible every 20 seconds
+	if needReconcileMiddleware {
+		r.Log.V(2).Info("ReconcileMiddleware", "message", message)
+		conditionErr := AddFailedCondition(ctx, r.Client, mgh, message)
+		if conditionErr != nil {
+			return ctrl.Result{}, conditionErr
+		}
+		return reconcile.Result{RequeueAfter: 20 * time.Second}, nil
```
clyang82, line 197: wait for 20 seconds is too long. I would like to keep 5 seconds.

operator/pkg/controllers/hubofhubs/globalhub_controller.go
```diff
@@ -167,12 +178,41 @@ func (r *MulticlusterGlobalHubReconciler) Reconcile(ctx context.Context, req ctr
 		return ctrl.Result{Requeue: true}, err
 	}
 
-	if result, err := r.ReconcileMiddleware(ctx, mgh); err != nil {
-		return result, err
+	needReconcileMiddleware, message, err := r.ReconcileMiddleware(ctx, mgh)
```
clyang82, line 181: How about keep the method no change, then AddFailedCondition at the beginning of `Reconcile()` after `if config.IsPaused(mgh) {`,
update the Condition is true at the end of `Reconcile`. What do you think?

clyang82 (approved):
Great enhancement. Thanks @ldpliu

Issue #685, opened 2023-10-18
Fixed the kafka versoin to 3.5.0


yanmxa:
<img width="1513" alt="image" src="https://github.com/stolostron/multicluster-global-hub/assets/19286664/e7ae3bc3-11ba-4f10-985c-36c7298fe65f">

PR #686, opened 2023-10-18
Fix the Kafka version to avoid bugs caused by version upgrades

yanmxa:
Resolved: https://github.com/stolostron/multicluster-global-hub/issues/685


morvencao (commented):
/lgtm

Issue #687, opened 2023-10-18
Get the DATABASE_URL from secret


clyang82:
We should not include the DATABASE_URL as plain text in the global hub manager deployment. the value should be from
1. if it is customized, the value is from the `multicluster-global-hub-storage` secret
2. if it provides by us, the value should be from `multicluster-global-hub-postgres` secret. Need add a field in this secret.

PR #690, opened 2023-10-18
Implicitly pass the postgres url to manager

yanmxa:
Resolved: https://github.com/stolostron/multicluster-global-hub/issues/687

Ref: https://issues.redhat.com/browse/ACM-8230

https://github.com/stolostron/multicluster-global-hub/issues/691: If we want to dynamically load the values of the Postgres credential into the pod, we need a separate controller to load and update the connection of the database, the same as Kafka.

### DIFF
--- a/operator/pkg/constants/constants.go
+++ b/operator/pkg/constants/constants.go
@@ -87,9 +87,10 @@ const (
 
 // global hub transport and storage secret names
 const (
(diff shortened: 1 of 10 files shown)


clyang82 (commented):
How do you resolve the problem that reload the postgres-credential if the ca or url is changed?

PR #692, opened 2023-10-19
add known issue

ldpliu:
https://issues.redhat.com/browse/ACM-6230
https://issues.redhat.com/browse/ACM-7461


clyang82:
maybe we also need to log a task for ID to add them into official document. what do you think?

ldpliu:
> maybe we also need to log a task for ID to add them into official document. what do you think?

Yes, created.
https://issues.redhat.com/browse/ACM-8264

clyang82:
Thanks

/lgtm
/approved

PR #697, opened 2023-10-23
fix post submit sonarcloud

yanmxa:

### DIFF
--- a/operator/pkg/controllers/hubofhubs/globalhub_controller.go
+++ b/operator/pkg/controllers/hubofhubs/globalhub_controller.go
@@ -161,6 +161,12 @@ func (r *MulticlusterGlobalHubReconciler) Reconcile(ctx context.Context, req ctr
 		return ctrl.Result{}, nil
 	}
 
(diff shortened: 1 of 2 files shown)


operator/pkg/controllers/hubofhubs/integration_test.go
```diff
@@ -955,29 +955,6 @@ var _ = Describe("MulticlusterGlobalHub controller", Ordered, func() {
...
-			Expect(k8sClient.Create(ctx, mcgh)).Should(Succeed())
-		})
-
-		It("Should get the postgres connection", func() {
-			mghReconciler.MiddlewareConfig.PgConnection = nil
-			err := mghReconciler.InitPostgresByStatefulset(ctx, mcgh)
-			if err != nil {
-				fmt.Println("InitPostgresBystatefuleset Error", err.Error())
-			}
-			Expect(mghReconciler.MiddlewareConfig.PgConnection).ShouldNot(BeNil())
```
clyang82, line 978: why remove this case? It is used to check if the initialize the postgres can be successful.

PR #703, opened 2023-10-26
Create Kafka certs for each managed hub

yanmxa:

Resolved: https://issues.redhat.com/browse/ACM-6179
Reference: https://github.com/strimzi/strimzi-kafka-operator/tree/main/documentation/modules/security

We will disable this feature for the BYO case.  Since we don't know which kind of Kafka cluster the user provisioned, whether cloud we create the connection credential by the `KafkaUser` API under the global hub namespace.

So If the customer provides the transport secret. We will use that as the only source of truth for connecting Kafka, whether from a global hub cluster or managed hub clusters.

Also, we aren't considering renewing the certificates for now.


clyang82:
Good job!
/lgtm

PR #710, opened 2023-11-06
Update the global hub 2.9 document

yanmxa:


doc/README.md
```diff
@@ -41,7 +41,7 @@ You can read about the use cases for multicluster global hub in [Use Cases](./gl
 
 ### Multicluster Global Hub Operator
 
-The Multicluster Global Hub Operator contains the components of multicluster global hub. The Operator deploys all of the required components for global multicluster management. The components include `multicluster-global-hub-manager` in the global hub cluster and `multicluster-global-hub-agent` in the managed hub clusters.
+The Multicluster Global Hub Operator contains the components of multicluster global hub. The Operator deploys all of the required components for global multicluster management. The components include `multicluster-global-hub-manager` and `multicluster-global-hub-grafana` in the global hub cluster and `multicluster-global-hub-agent` in the managed hub clusters.
```
clyang82, line 44: Do we need to mention postgres and kafka?

doc/README.md
```diff
@@ -294,18 +312,23 @@ After installing the global hub operand, the global hub manager starts running a
 
 #### Local compliance status sync job
 
-  At 0 o'clock every day, based on the policy status and events collected by the manager on the previous day. Running the job to summarize the compliance status and change frequency of the policy on the cluster, and store them to the `history.local_compliance` table as the data source of grafana dashboards. Please refer to [here](./how_global_hub_works.md) for more details.
+  At 0 o'clock every day, based on the policy status and events collected by the manager on the previous day. Running the job to summarize the compliance status and change frequency of the policy on the cluster, and store them to the `history.local_compliance` table as the data source of grafana dashboards. Even if the job runs once a day, it can also run multiple times within the day without changing the result. Please refer to [here](./how_global_hub_works.md) for more details.
```
clyang82, line 315: do not understand `Even if the job runs once a day, it can also run multiple times within the day without changing the result.`

yanmxa: That means if we run the job several times in one day. the result will be the same.

doc/README.md
yanmxa: Yes. Since the operator will do such a thing, let's add this information.

Done!

doc/README.md
yanmxa: @clyang82 Maybe we can remove it from here?

clyang82: yes. just remove to avoid some confusions.

yanmxa: @clyang82 Done.

PR #713, opened 2023-11-09
Trace the error log location

yanmxa:
resolved: https://github.com/stolostron/multicluster-global-hub/issues/607

- Format: add the color identification for different log level
- Trace the error log location when catch the error log

Example:
image


clyang82:
Do not need to cherry-pick to release branch. Let me have it in main branch only.

/lgtm

PR #715, opened 2023-11-10
add leaf hub id

ldpliu:
https://issues.redhat.com/browse/ACM-8409

### DIFF
--- a/agent/pkg/status/controller/hubcluster/clusters_status_sync.go
+++ b/agent/pkg/status/controller/hubcluster/clusters_status_sync.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	routev1 "github.com/openshift/api/route/v1"
+	clustersv1alpha1 "open-cluster-management.io/api/cluster/v1alpha1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
@@ -30,12 +31,31 @@ func AddHubClusterInfoSyncer(mgr ctrl.Manager, producer transport.Producer) erro
 	} // bundle predicate - always send subscription status.
 
 	hubClusterInfoPredicate := predicate.NewPredicateFuncs(func(object client.Object) bool {
-		return (object.GetNamespace() == constants.OpenShiftConsoleNamespace &&
-			object.GetName() == constants.OpenShiftConsoleRouteName) ||
-			(object.GetNamespace() == constants.ObservabilityNamespace &&
-				object.GetName() == constants.ObservabilityGrafanaRouteName)
+		if object.GetNamespace() == constants.OpenShiftConsoleNamespace &&
+			object.GetName() == constants.OpenShiftConsoleRouteName {
+			return true
+		}
+		if object.GetNamespace() == constants.ObservabilityNamespace &&
+			object.GetName() == constants.ObservabilityGrafanaRouteName {
+			return true
+		}
+		return false
 	})
 
-	return generic.NewGenericStatusSyncer(mgr, "hub-cluster-status-sync", producer, bundleCollection,
+	err := generic.NewGenericStatusSyncer(mgr, "hub-cluster-status-sync", producer, bundleCollection,
 		createObjFunction, hubClusterInfoPredicate, config.GetHubClusterInfoDuration)
+	if err != nil {
+		return err
+	}
+
+	createClaimObjFunction := func() bundle.Object { return &clustersv1alpha1.ClusterClaim{} }
+	hubClusterClaimPredicate := predicate.NewPredicateFuncs(func(object client.Object) bool {
+		return object.GetName() == "id.k8s.io"
+	})
+	err = generic.NewGenericStatusSyncer(mgr, "hub-cluster-status-sync", producer, bundleCollection,
+		createClaimObjFunction, hubClusterClaimPredicate, config.GetHubClusterInfoDuration)
+	if err != nil {
+		return err
+	}
+	return nil
 }
--- a/manager/pkg/statussyncer/syncers/hub_cluster_info_syncer.go
+++ b/manager/pkg/statussyncer/syncers/hub_cluster_info_syncer.go
@@ -3,10 +3,8 @@ package dbsyncer
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 
 	"github.com/go-logr/logr"
-	"gorm.io/gorm/clause"
 
 	"github.com/stolostron/multicluster-global-hub/pkg/bundle"
 	"github.com/stolostron/multicluster-global-hub/pkg/bundle/base"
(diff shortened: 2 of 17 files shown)


agent/pkg/status/controller/hubcluster/clusters_status_sync.go
```diff
@@ -64,28 +65,36 @@ func (c *hubClusterController) Reconcile(ctx context.Context, request ctrl.Reque
...
+
+	c.syncBundle(ctx)
 
 	reqLogger.V(2).Info("Reconciliation complete.")
 
 	return ctrl.Result{}, nil
 }
 
-func (c *hubClusterController) syncBundle(ctx context.Context, route *routev1.Route) {
-	c.bundle.UpdateObject(route)
```
clyang82, line 87: what is purpose to move it out of this syncBundle method?

operator/pkg/controllers/hubofhubs/globalhub_database.go
```diff
@@ -74,6 +82,27 @@ func (r *MulticlusterGlobalHubReconciler) ReconcileDatabase(ctx context.Context,
 	return nil
 }
 
+func handleUpgrade(ctx context.Context, conn *pgx.Conn, log logr.Logger, upgradeFS embed.FS, username string) error {
```
clyang82, line 85: I am thinking a case that if the table is altered, but the global hub agent is still old. so it can send the data without cluster_id. it reports the error which can generate lot of logs.

clyang82 (commented):
Do we need to add `cluster_id` in event.local_policies table? otherwise, how to fix the problem in https://issues.redhat.com/browse/ACM-8409

clyang82:
@ldpliu that is a big changes. Could you write down a proposal or a solution? Thanks.

ldpliu:
@clyang82 @yanmxa 
I have changed the pr follow our discussion. Please help to take a look again.
<Will handle failed case in another pr>

pkg/constants/constants.go
```diff
@@ -22,6 +22,8 @@ const (
 
 	DefaultAgentQPS   = float32(150)
 	DefaultAgentBurst = 300
+
+	DefaultClusterId = "00000000-0000-0000-0000-000000000000"
```
yanmxa, line 26: Since we using a fake ID for the existing cluster, how about generating a unique ID and make it the primary key for `the status.leaf_hub`, instead the joint primary key?

ldpliu: As we need to known the default id when we update the table. It is also used in dashboard part, So I think current id is necessary.

yanmxa: I was thinking maybe we could set an id with a value like `prefix + random` to indicate it is just a placeholder and then make the id column the primary key.
So your solution is to work in such a case. But I wonder If the ID isn't unique, It may cause other issues when we aggregate the data by ID. 

The initial purpose of introducing the `id` is because the `leafhub_name` mightn't be unique for the cluster.  But we add the `ID` also can be conflicted with each other. It seems a little bit strange.

Anyway the can work for the current scenario. We can merge this if you insist.

manager/pkg/statussyncer/syncers/hub_cluster_info_db_syncer.go
```diff
@@ -81,47 +84,70 @@ func (syncer *hubClusterInfoDBSyncer) handleLocalObjectsBundle(ctx context.Conte
...
-
-	batchLeafhub := []models.LeafHub{}
 	for _, object := range bundle.GetObjects() {
+		clusterId := constants.DefaultClusterId
 		specificObj, ok := object.(*status.LeafHubClusterInfo)
 		if !ok {
 			continue
 		}
+		mc := clusterv1.ManagedCluster{}
+		err := (*syncer.client).Get(ctx, types.NamespacedName{
```
clyang82, line 95: I think it is better to add the clusterID from the global hub agent. We can get it from the clusterclaim CR. It is unnecessary to get the clusterclaim from the managed cluster in the global hub side. WDYT?

go.mod
```diff
@@ -62,7 +61,10 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
-require sigs.k8s.io/kube-storage-version-migrator v0.0.6-0.20230721195810-5c8923c5ff96 // indirect
+require (
+	github.com/jackc/pgconn v1.12.1 // indirect
+	sigs.k8s.io/kube-storage-version-migrator v0.0.6-0.20230721195810-5c8923c5ff96 // indirect
+)
```
clyang82, line 68: unnecessary change?

manager/pkg/statussyncer/syncers/hub_cluster_info_syncer.go
```diff
@@ -74,38 +72,58 @@ func (syncer *hubClusterInfoDBSyncer) handleLocalObjectsBundle(ctx context.Conte
...
-			Payload:     payload,
-		})
-	}
-
-	err = db.Clauses(clause.OnConflict{
-		UpdateAll: true,
-	}).CreateInBatches(batchLeafHub, 100).Error
-	if err != nil {
-		return err
+		err = db.Where("leaf_hub_name = ?", leafHubName).Find(&existingObjs).Error
```
clyang82, line 91: move this statement out of the for loop. it is nothing related with for loop.

manager/pkg/statussyncer/syncers/hub_cluster_info_syncer.go
```diff
@@ -74,38 +72,58 @@ func (syncer *hubClusterInfoDBSyncer) handleLocalObjectsBundle(ctx context.Conte
...
+					Or(&models.LeafHub{
+						LeafHubName: leafHubName,
+						ClusterID:   clusterId,
+					}).
+					Updates(&models.LeafHub{
+						LeafHubName: leafHubName,
+						ClusterID:   clusterId,
+						Payload:     payload,
+					}).Error
+				if err != nil {
```
clyang82, line 122: can we continue to use the previous logic to update the db in batch?

ldpliu:
> correct me if i am wrong. in your code, it is possible to insert the default clusterId `00000000-0000-0000-0000-000000000000`. since the clusterid + leaf hub name is primary key, how do you update it?

In normal case, the clusterid should not be `00000000-0000-0000-0000-000000000000`. It only used to handle upgrade case(destroyed clusters in status.leaf_hubs) or cluster claim not found case. For these cases which use the default clusterid, we use the leafhub name to identify the unique leafhub.

manager/pkg/statussyncer/syncers/hub_cluster_info_syncer.go
ldpliu: > can we continue to use the previous logic to update the db in batch?

Actually, it's difficult to handle the logic use batch job.

And in the [bundle](https://github.com/stolostron/multicluster-global-hub/pull/715/files#diff-d777344f3d335e08ad16334adc3dc9471df546bc25ff446c3e1fb8d0a106ec94R76), it only have one item, So I have changed the logic and it only update db once also.

yanmxa:
Reference: The deletion https://github.com/stolostron/multicluster-global-hub/blob/main/manager/pkg/specsyncer/spec2db/controller/managedhub_spec_sync.go

manager/pkg/statussyncer/syncers/hub_cluster_info_syncer.go
```diff
@@ -74,38 +72,58 @@ func (syncer *hubClusterInfoDBSyncer) handleLocalObjectsBundle(ctx context.Conte
...
+			err := db.Create(&models.LeafHub{
+				LeafHubName: leafHubName,
+				ClusterID:   clusterId,
+				Payload:     payload,
+			}).Error
+			return err
+		}
+		err = db.Model(&models.LeafHub{}).
+			Where(&models.LeafHub{
+				LeafHubName: leafHubName,
```
clyang82, line 117: Add `deleted_at` is NULL as a condition.

yanmxa:
/lgtm 

once it is merged, I need to consolidate by this: https://github.com/stolostron/multicluster-global-hub/pull/728

PR #717, opened 2023-11-13
Consolidate the bundles between manager and agent

yanmxa:
- Combine the manager and agent bundle into one entity
- Abstract the bundle interface to reduce the difficulty of the extensibility 
- Refactor the Conflation with the enhanced bundle to make it easier to read

PR #721, opened 2023-11-21
Replace the sql builder with ORM(Object-Relational Mapping)

yanmxa:

1. Deprecate the pgx connection pool(fixed connections) with gorm database connection pool(dynamic db connection size)
2. Remove the poor readability SQL builder from the database
3. Remove the heartbeat bundle for the managed hub 

This is split from PR: https://github.com/stolostron/multicluster-global-hub/pull/717


agent/pkg/status/controller/local_policies/local_replicas_policy_sync.go
```diff
@@ -11,37 +11,32 @@ import (
...
-	agentstatusconfig "github.com/stolostron/multicluster-global-hub/agent/pkg/status/controller/config"
+	"github.com/stolostron/multicluster-global-hub/agent/pkg/status/controller/config"
 	"github.com/stolostron/multicluster-global-hub/agent/pkg/status/controller/generic"
 	"github.com/stolostron/multicluster-global-hub/pkg/constants"
 	"github.com/stolostron/multicluster-global-hub/pkg/transport"
 )
 
 // AddLocalPoliciesController this function adds a new local policies sync controller.
-func AddLocalClusterPolicyEventsController(mgr ctrl.Manager, producer transport.Producer) error {
+func AddLocalReplicasPoliciesSyncer(mgr ctrl.Manager, producer transport.Producer) error {
```
clyang82, line 21: why propose to call `AddLocalReplicasPoliciesSyncer`? what does that mean? the controller is for policy events, is not it?

clyang82 (commented):
It looks good for me.

Do you think we should enable a unit-test for all? I mean include the global-resources test

agent/pkg/status/controller/local_policies/local_replicas_policy_sync.go
yanmxa: There are two types of syncers for the policy. One is `LocalRootPolicySyncer`, which aims to report the root policy status, like compliance. The other is `LocalReplicasPolicySyncer`, it is used to report the status of replicas policy, like the policy history event, so it may include other replicas' information in the future.

yanmxa:
> It looks good for me.
> 
> Do you think we should enable a unit-test for all? I mean include the global-resources test

Since we care about the global resource. I think we can open the e2e test for all.

agent/pkg/status/controller/local_policies/local_replicas_policy_sync.go
clyang82: then we should call `LocalReplicatedPolicySyncer`.

PR #723, opened 2023-11-23
Refactor the transport with different protocol

yanmxa:

ref: https://issues.redhat.com/browse/ACM-8793

We define the transporter interface as the source of truth for different implementations. 

```go
// init the transport with different implementation/protocol: secret, strimzi operator or plain deployment
type Transporter interface {
	// define the user for each hub clusters, user lifecycle is managed by the transporter
        GenerateUserName(clusterIdentity string) string                       // "kafka-user-clusterIdentity"
	CreateUser(name string) error
	DeleteUser(name string) error

	// define the topic for each hub clusters(or shared topic), topic lifecycle is managed by the transporter
	GenerateClusterTopic(clusterIdentity string) *ClusterTopic        // (spec/status/event)
	CreateClusterTopic(topic *ClusterTopic ) error    
	DeleteClusterTopic(topic *ClusterTopic ) error

	// get the connection credential by user
	GetConnCredential(userName string) (*ConnCredential, error)
}
```


operator/pkg/controllers/hubofhubs/globalhub_middleware.go
```diff
@@ -0,0 +1,208 @@
...
+		}, r.Client)
+	}
+
+	// create the user to connect the transport instance
+	err = trans.CreateUser(transportprotocol.DefaultGlobalHubKafkaUser)
+	if err != nil {
+		return nil, err
+	}
+	// create global hub topics
+	topics := trans.GenerateClusterTopic("")
```
clyang82, line 120: Why we need to create topics here?

operator/pkg/controllers/hubofhubs/globalhub_middleware.go
```diff
@@ -0,0 +1,208 @@
...
+		}
+	case transport.SecretTransporter:
+		trans = transportprotocol.NewSecretTransporter(ctx, types.NamespacedName{
+			Namespace: mgh.Namespace,
+			Name:      constants.GHTransportSecretName,
+		}, r.Client)
+	}
+
+	// create the user to connect the transport instance
+	err = trans.CreateUser(transportprotocol.DefaultGlobalHubKafkaUser)
```
clyang82, line 115: Do you want to treat this user as a super user?

pkg/transport/transporter/secret_transporter.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/go-logr/logr"
+	"github.com/stolostron/multicluster-global-hub/pkg/transport"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type secretTransporter struct {
```
clyang82, line 16: it is not a type of transporter. if you want to define a type, maybe `BYOTransporter` make more sense.

operator/pkg/controllers/hubofhubs/globalhub_manager.go
```diff
@@ -188,3 +174,36 @@ func (r *MulticlusterGlobalHubReconciler) manipulateObj(ctx context.Context, hoh
 
 	return nil
 }
+
+type ManagerVariables struct {
```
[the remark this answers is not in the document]
yanmxa: The struct is also used in the test case, a common unit test error is that a field is inconsistent between the controller and the test case.

operator/pkg/controllers/hubofhubs/globalhub_middleware.go
yanmxa: It doesn't need to be super used. This user is just for the global hub manager to connect the transport

operator/pkg/controllers/hubofhubs/globalhub_middleware.go
yanmxa: It is to make sure the topic is created before the manager is deployed.

pkg/transport/transporter/secret_transporter.go
yanmxa: @clyang82 Done!

clyang82 (approved):
/lgtm

PR #724, opened 2023-11-27
add leaf hub id (no primary key change)

ldpliu:
https://issues.redhat.com/browse/ACM-8409


ldpliu:
In this pr, we do not change the primary key of status.leaf_hubs table.
And it could also fix the issue https://issues.redhat.com/browse/ACM-8409

Related pr: https://github.com/stolostron/multicluster-global-hub/pull/715

For dashboard, it will not have any impacts.

Difference:
As we use the leafhub name as the primary key, so it only has one item when user re-import a different hub with same name(we will only update the existing item if leafhub name exist).

ldpliu:
/cc @clyang82 @yanmxa This pr has no primary key change, please help to review.

https://github.com/stolostron/multicluster-global-hub/pull/724#issuecomment-1827435522

operator/pkg/controllers/hubofhubs/database/2.tables.sql
```diff
@@ -45,6 +45,7 @@ CREATE INDEX IF NOT EXISTS leafhub_cluster_idx ON status.managed_clusters (leaf_
 
 CREATE TABLE IF NOT EXISTS status.leaf_hubs (
     leaf_hub_name character varying(254) NOT NULL PRIMARY KEY,
+    cluster_id uuid NOT NULL DEFAULT '00000000-0000-0000-0000-000000000000',
     payload jsonb NOT NULL,
     console_url text generated always as (payload ->> 'consoleURL') stored,
     grafana_url text generated always as (payload ->> 'grafanaURL') stored,
```
clyang82, line 51: How do we have two records with the same leaf hub name in this table if the leaf_hub_name is primary key?

PR #725, opened 2023-11-27
Retrying insert bundle into with a threshold

yanmxa:

Add a new bundle status implementation that supports retry handling.

ref: https://github.com/stolostron/multicluster-global-hub/issues/729


ldpliu:
When the bundle process failed, we need to wait some seconds(30s) then process it next time.

yanmxa:
> When the bundle process failed, we need to wait some seconds(30s) then process it next time.

The current process queue is not suitable for the deploy case, if it is a very important scenario. we should re-design its mechanism. Cause it needs to change a lot.

Issue #726, opened 2023-11-28
Request to add multi-kube objects for a shared bundle


yanmxa:
Currently, we can attach multi-bundle to an object, like `compliance` and `completeCompliance` bundle for root policy. 
Then the agent leverage the generic syncer to load information to the bundle. 

However, we can't support handling the `n-1` cases(n means more than 1 object). I suggest extending another `shared-bundle-syncer` that can be used to support multiple objects to update a bundle.

In this way, our bundles can be divided into two types:

- exclusive bundle: all the current global hub bundles, it's processed by the `generic-syncer`
- shared bundle: we may aggregate more than one object resource into one bundle, that can be processed by the `multi-handler-bundle-sycner`

## Object Vs Bundle

- `1 : n` -  `policy` -> `complianceBundle / completeComplianceBundle, deltabundle`
- `n : 1` -  `Route/ ClusterClaim` -> `HubClusterInfoBundle`

- The drawbacks to using `GenericStatusSyncer` sync `n : 1` bundle
  - the bundle process logic will become redundant and hard to maintain
  - result in the performance degradation of the manager
  
<img width="981" alt="image" src="https://github.com/stolostron/multicluster-global-hub/assets/19286664/dfbdcd2e-51d2-4380-8dfd-4a325c759995">

- Propose adding `MultiHandlerStatusSyncer` to handle `n : 1` bundle 
<img width="337" alt="image" src="https://github.com/stolostron/multicluster-global-hub/assets/19286664/8a9dc86c-3587-49b3-9225-fd3334d5b13b">

yanmxa:
object - bundle:  
 - 1 -> n: policy -> complianceBundle / completeComplianceBundle, deltabundle        => bundleCollection
 - n -> 1

PR #728, opened 2023-11-28
Add shared bundle syncer for multi-kub objects

yanmxa:

resolved: https://github.com/stolostron/multicluster-global-hub/issues/726#issue-2013889807
reference: https://github.com/stolostron/multicluster-global-hub/pull/715


clyang82:
That is a great idea to have multiple kube resources tackling with one syncer. 👍 always keep in mind on the resource consumption

/lgtm

PR #731, opened 2023-12-01
resend hub info bundle when manager start

ldpliu:
https://issues.redhat.com/browse/ACM-9179
In upgrade case, there is a middle status that agent upgrade to new version, but manager do not. 
When agent start, agent will send the managed hubinfo(we add clusterid in this release) to manager, but manager can not handle it now and will discard it. 
After manager upgrade to new version, agent will not send the hubinfo any more, so the hubinfo in database could not be updated. So we need to ask agent to resend hubinfo after manager started.

### DIFF
--- a/agent/pkg/controllers/crd_controller.go
+++ b/agent/pkg/controllers/crd_controller.go
@@ -33,12 +33,10 @@ func (c *crdController) Reconcile(ctx context.Context, request ctrl.Request) (ct
 	reqLogger.V(2).Info("crd controller", "NamespacedName:", request.NamespacedName)
 
 	// add spec controllers
-	if c.agentConfig.EnableGlobalResource {
-		if err := specController.AddToManager(c.mgr, c.agentConfig); err != nil {
-			return ctrl.Result{}, fmt.Errorf("failed to add spec syncer: %w", err)
-		}
-		reqLogger.V(2).Info("add spec controllers to manager")
+	if err := specController.AddToManager(c.mgr, c.agentConfig); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to add spec syncer: %w", err)
 	}
+	reqLogger.V(2).Info("add spec controllers to manager")
 
 	if err := statusController.AddControllers(ctx, c.mgr, c.agentConfig); err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to add status syncer: %w", err)
--- a/agent/pkg/spec/controller/syncers/resend_syncer.go
+++ b/agent/pkg/spec/controller/syncers/resend_syncer.go
@@ -1,0 +1,44 @@
+package syncers
+
+import (
+	"encoding/json"
+
+	"github.com/go-logr/logr"
+	"github.com/stolostron/multicluster-global-hub/agent/pkg/status/controller/cache"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+// resyncSyncer resync the bundle info.
+type resyncSyncer struct {
+	log logr.Logger
+}
+
+func NewResyncSyncer() *resyncSyncer {
+	return &resyncSyncer{
+		log: ctrl.Log.WithName("Resync status syncer"),
+	}
+}
+
+func (syncer *resyncSyncer) Sync(payload []byte) error {
+	syncer.log.Info("Resync bundle info")
+	bundleKeys := []string{}
+	if err := json.Unmarshal(payload, &bundleKeys); err != nil {
+		syncer.log.Error(err, "Failed to unmarshal bundle keys")
+		return err
+	}
+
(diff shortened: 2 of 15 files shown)


pkg/bundle/cluster/hub_cluster_info_bundle.go
```diff
@@ -20,7 +20,7 @@ type HubClusterInfoBundle struct {
 }
 
 // LeafHubClusterInfoStatusBundle creates a new instance of LeafHubClusterInfoStatusBundle.
-func NewAgentHubClusterInfoBundle(leafHubName string) bundle.BaseAgentBundle {
```
yanmxa, line 23: This logic is about resync the bundle from agent to manager, why do we update the hub cluster info bundle?

ldpliu: > This logic is about resync the bundle from agent to manager, why do we update the hub cluster info bundle?

Because we need to add the increase bundle version  function:
https://github.com/stolostron/multicluster-global-hub/pull/731/files#diff-4c54f3148cd2013f948426697c71c651cf618299b4af12cf5d710644015ab76aR41

agent/pkg/spec/controller/syncers/resend_syncer.go
```diff
@@ -0,0 +1,41 @@
...
+	bundleKeys := []string{}
+	if err := json.Unmarshal(payload, &bundleKeys); err != nil {
+		syncer.log.Error(err, "Failed to unmarshal bundle keys")
+		return err
+	}
+
+	for _, bundleKey := range bundleKeys {
+		syncer.log.Info("Resend bundle", "key", bundleKey)
+		if bundleKey == constants.HubClusterInfoMsgKey {
+			hubcluster.IncreaseBundleVersion()
```
yanmxa, line 36: Use BundleVersion to incr the version. We don't need to extend the bundle to update the version.

yanmxa: Also maybe we should provide a generic way to fetch the bundle by transport MsgKey. Otherwise, maintaining backward compatibility will become very redundant.

agent/pkg/status/controller/cache/cache.go
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) 2024 Red Hat, Inc.
```
yanmxa, line 1: Using cache to retrieve the specific bundle is okay!

Also, I'm thinking of another way not to cache the bundle:
Since all the syncers hold the bundle, and all the syncers have been registered to the dispatcher. So we can extend the dispatcher's ability to retrieve the bundle with the key.

yanmxa:
The overall structure looks good to me. Leave it to @clyang82!

agent/pkg/spec/controller/syncers/resend_syncer.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/go-logr/logr"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/stolostron/multicluster-global-hub/agent/pkg/status/controller/hubcluster"
+
+	"github.com/stolostron/multicluster-global-hub/pkg/constants"
+)
+
+// managedClusterLabelsBundleSyncer syncs managed clusters metadata from received bundles.
```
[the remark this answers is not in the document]
ldpliu: Done

agent/pkg/controllers/crd_controller.go
```diff
@@ -33,12 +33,10 @@ func (c *crdController) Reconcile(ctx context.Context, request ctrl.Request) (ct
 	reqLogger.V(2).Info("crd controller", "NamespacedName:", request.NamespacedName)
 
 	// add spec controllers
-	if c.agentConfig.EnableGlobalResource {
```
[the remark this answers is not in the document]
ldpliu: We need to use the logic [here](https://github.com/stolostron/multicluster-global-hub/pull/731/files#diff-b94affa97b961fd4b201fdeb76405ac7c608eea1ca78b50edf9a9892f83b0f4bR15), and we also check the enable global resource [here](https://github.com/stolostron/multicluster-global-hub/pull/731/files#diff-b94affa97b961fd4b201fdeb76405ac7c608eea1ca78b50edf9a9892f83b0f4bR38)

pkg/constants/constants.go
```diff
@@ -102,6 +102,8 @@ const (
 	// HubClusterInfoMsgKey - hub cluster info message key.
 	HubClusterInfoMsgKey = "HubClusterInfo"
 
+	ResendMsgKey = "ResendInfo"
```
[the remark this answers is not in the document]
ldpliu: done

PR #732, opened 2023-12-01
restart manager when middleware change

ldpliu:
https://issues.redhat.com/browse/ACM-7489

### DIFF
--- a/operator/pkg/controllers/hubofhubs/globalhub_manager.go
+++ b/operator/pkg/controllers/hubofhubs/globalhub_manager.go
@@ -4,14 +4,18 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"reflect"
 	"strconv"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/discovery/cached/memory"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/restmapper"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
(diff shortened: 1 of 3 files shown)


operator/pkg/controllers/hubofhubs/globalhub_manager.go
```diff
@@ -20,12 +24,19 @@ import (
...
 	"github.com/stolostron/multicluster-global-hub/operator/pkg/renderer"
 	"github.com/stolostron/multicluster-global-hub/pkg/constants"
 	"github.com/stolostron/multicluster-global-hub/pkg/transport"
 	commonutils "github.com/stolostron/multicluster-global-hub/pkg/utils"
 )
 
+var pgConnectionCache *postgres.PostgresConnection
+var kafkaConnectionCache *kafka.KafkaConnection
+
+const managerDeploymentName = "multicluster-global-hub-manager"
```
clyang82, line 38: suggest to move it into constants.go file

operator/pkg/controllers/hubofhubs/globalhub_manager.go
```diff
@@ -78,6 +89,12 @@ func (r *MulticlusterGlobalHubReconciler) reconcileManager(ctx context.Context,
 		replicas = 2
 	}
 
+	if isMiddleWareChanged(r.MiddlewareConfig) {
```
clyang82, line 92: middleware is a word. let us make `W` as lower case

operator/pkg/controllers/hubofhubs/globalhub_manager.go
```diff
@@ -151,6 +168,64 @@ func (r *MulticlusterGlobalHubReconciler) reconcileManager(ctx context.Context,
 	return nil
 }
 
+func restartManagerPod(ctx context.Context, kubeClient kubernetes.Interface) error {
```
clyang82, line 171: restart the pod will lose the logs. can we just restart the kafkaConsumer and kafkaProducer? I remembered gorm can handle it automatically @yanmxa is that true?

operator/pkg/controllers/hubofhubs/globalhub_manager.go
ldpliu: Done

operator/pkg/controllers/hubofhubs/globalhub_manager.go
ldpliu: Done

operator/pkg/controllers/hubofhubs/globalhub_manager.go
yanmxa: @clyang82 @ldpliu Sorry for not replying in time. 

- The gorm can't reload/reconnect the credential automatically for now.
- I had several ideas for propagating the middleware credentials, and before we choose the restarting manager, There are the following options we should consider(both of them don't need to restart the pod/manager):
  1. The operator calls the non-k8s API of the manager to update their credential. Then the manager rebuilds the connection to these middlewares.
      As far as I know, a lot of services provide a **refreshToken** API dedicated to updating the expired credentials. So I think we can prioritize this approach.
  2. The global hub manager watches the credentials of these middlewares, and rebuild the connections in the controller.

PR #739, opened 2023-12-15
Agent should be running without ACM hub installed

clyang82:
the PR address the following cases:
1. if the managed cluster is not a managed hub cluster, the agent should be running normally w/o restarting. And the `hub.open-cluster-management.io` clusterclaim value is `NotInstalled`
2. if add label `global-hub.open-cluster-management.io/hub-cluster-install=` to this managed cluster, the agent should keep running and start the corresponding controllers to handle the policy and managedcluster.

How to address the above cases:
1. w/o MCH or clustermanager, start the controller to create `hub.open-cluster-management.io` clusterclaim and start the crd controller to watch the crd.
2. if the crd `clustermanager` is created, then start the rest of controllers.

ref: https://issues.redhat.com/browse/ACM-8556


agent/pkg/controllers/crd_controller.go
```diff
@@ -0,0 +1,77 @@
...
+// if the crd exists, then add controllers to the manager dynamically
+func StartCRDController(mgr ctrl.Manager, restConfig *rest.Config, agentConfig *config.AgentConfig) error {
+	crdPredicate := predicate.NewPredicateFuncs(func(object client.Object) bool {
+		if object.GetName() == "multiclusterhubs.operator.open-cluster-management.io" ||
+			object.GetName() == "clustermanagers.operator.open-cluster-management.io" {
+			return true
+		}
+		return false
+	})
+	return ctrl.NewControllerManagedBy(mgr).
```
yanmxa, line 69: Also, to detect these two CRDs, Do we need to watch(cache) all the CRDs? Here is another option - periodically check the target CRDs with a goroutine. If it finds them, then start the related controllers and exit the goroutine.

agent/pkg/controllers/crd_controller.go
```diff
@@ -0,0 +1,77 @@
...
+
+	// add spec controllers
+	if c.agentConfig.EnableGlobalResource {
+		if err := specController.AddToManager(c.mgr, c.agentConfig); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to add spec syncer: %w", err)
+		}
+		reqLogger.V(2).Info("add spec controllers to manager")
+	}
+
+	if err := statusController.AddControllers(ctx, c.mgr, c.agentConfig); err != nil {
```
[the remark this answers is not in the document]
clyang82: the manager supports add controllers in a dynamic way. refer to https://github.com/kubernetes-sigs/controller-runtime/issues/1994

agent/pkg/controllers/crd_controller.go
clyang82: Good point to saving the memory consumption. instead of using goroutine to check periodically, I prefer to limit the cache by using https://github.com/kubernetes-sigs/controller-runtime/blob/main/designs/use-selectors-at-cache.md

yanmxa: Another reason is that when the CRD controller watches the target CRD is created. The CRD controller should be deprecated. So if you insist on using the controller, maybe we can move the logic to the `start` instead of `reconcile`， because it's a one-off deal.

agent/pkg/scheme/scheme.go
```diff
@@ -1,14 +1,15 @@
-// Copyright (c) 2022 Red Hat, Inc.
+// Copyright (c) 2022 Red Hat(scheme)) Inc.
```
yanmxa, line 1: -（scheme）?
- s/2022/2023

agent/pkg/controllers/crd_controller.go
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) 2022 Red Hat, Inc.
```
yanmxa, line 1: 2023

PR #740, opened 2023-12-15
add backup labels

ldpliu:
Story: https://issues.redhat.com/browse/ACM-8873

Add `cluster.open-cluster-management.io/backup:globalhub` label for the following resources in **Global hub** namespace:
1. Customize secrets
    - "multicluster-global-hub-custom-grafana-config"
    - "multicluster-global-hub-transport"
    - "multicluster-global-hub-storage"
2. Customize configmap
    - multicluster-global-hub-custom-alerting
4. kafka cluster
5. all kafka users
6. All kafka topic
7. kafka crds
    - "kafkas.kafka.strimzi.io"
    -  "kafkatopics.kafka.strimzi.io"
    - "kafkausers.kafka.strimzi.io"

Add `cluster.open-cluster-management.io/backup:cluster-activation` label for **mgh** instance in **Global hub** namespace

Add `cluster.open-cluster-management.io/volsync:globalhub` label for all PVC in **Global hub** namespace


operator/pkg/controllers/backup/crd.go
```diff
@@ -0,0 +1,68 @@
+/*
```
clyang82, line 1: is it possible to merge different type of resources into a general method? something like https://github.com/stolostron/multicluster-global-hub/blob/b23c6fc5756ae6938847761da951ddca5e1018de/operator/pkg/deployer/hoh_deployer.go#L32

because what we need to handle is to add label for the resources, we do not need to translate to concrete object, maybe &unstructured.Unstructured{} is enough. WDYT?

clyang82 (approved):
looks good for me.

@ldpliu please have another PR to detect the backup & restore is enabled or not to decide if we need to start the global hub backup or not.

PR #744, opened 2023-12-20
Split topics for each regional hub

yanmxa:
Ref: https://issues.redhat.com/browse/ACM-8793

image

If we split the spec, then the manager must get which cluster it needs to send the message. That means it has to hold the cluster information

Note:
- e2e certificate error reference: https://github.com/confluentinc/librdkafka/issues/4349
	- resolved by setting: "ssl.endpoint.identification.algorithm": "none",
```bash
%3|1703161332.046|FAIL|multicluster-global-hub-manager#consumer-2| [thrd:ssl://172.20.0.2:30095/bootstrap]: ssl://172.20.0.2:30095/bootstrap: SSL handshake failed: error:0A000086:SSL routines::certificate verify failed: broker certificate could not be verified, verify that ssl.ca.location is correctly configured or root CA certificates are installed (install ca-certificates package) (after 7ms in state SSL_HANDSHAKE, 30 identical error(s) suppressed)
```
- override the UT test coverage, cause the codes mostly is about the cloudevents implementations

### DIFF
--- a/operator/pkg/transporter/strimzi_transporter.go
+++ b/operator/pkg/transporter/strimzi_transporter.go
@@ -313,6 +313,13 @@ func (k *strimziTransporter) GetConnCredential(username string) (*transport.Conn
 }
 
 func (k *strimziTransporter) newKafkaTopic(topicName string) *kafkav1beta2.KafkaTopic {
+	replicas := DefaultReplicas
+	// if tls is not enabled(that means its the dev environment), the replicas must be 1
+	// of it may occur such error 'Replication factor: 2 larger than available brokers: 1.'
+	if !k.enableTLS {
+		replicas = 1
+	}
+
 	return &kafkav1beta2.KafkaTopic{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      topicName,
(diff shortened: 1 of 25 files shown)


operator/pkg/transporter/strimzi_transporter.go
```diff
@@ -219,21 +220,28 @@ func (k *strimziTransporter) DeleteUser(topicName string) error {
...
+		}
+	}
+
+	return topic
 }
 
 func (k *strimziTransporter) CreateTopic(topic *transport.ClusterTopic) error {
 	for _, topicName := range []string{topic.SpecTopic, topic.StatusTopic, topic.EventTopic} {
+		// if the topicName = "^status.*", convert it to status.global for creating
+		if topicName == GlobalRegexStatusTopic {
```
clyang82, line 242: I am thinking that can we postpone the status.* subscription until the topics exist? in this way, we do not need to create `status.global` topic. You can address it in a separated PR.

clyang82 (approved):
/lgtm

clyang82:
/override "SonarCloud Code Analysis"

cannot pass the code coverage due to confluence part. we will move this part into ocm/sdk soon.

PR #749, opened 2023-12-23
Support resources

clyang82:
ref: https://issues.redhat.com/browse/ACM-9140

### DIFF
--- a/operator/apis/v1alpha4/multiclusterglobalhub_types.go
+++ b/operator/apis/v1alpha4/multiclusterglobalhub_types.go
@@ -90,6 +90,41 @@ type MulticlusterGlobalHubSpec struct {
 	// +kubebuilder:default={postgres: {retention: "18m"}}
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	DataLayer DataLayerConfig `json:"dataLayer"`
+	// Advanced configurations for global hub
+	// +optional
+	AdvancedConfig *AdvancedConfig `json:"advanced,omitempty"`
+}
+
+type AdvancedConfig struct {
+	// The spec of grafana
+	// +optional
+	Grafana *CommonSpec `json:"grafana,omitempty"`
+
+	// The spec of kafka
+	// +optional
+	Kafka *CommonSpec `json:"kafka,omitempty"`
+
+	// The spec of zookeeper
+	// +optional
+	Zookeeper *CommonSpec `json:"zookeeper,omitempty"`
+
+	// The spec of postgres
+	// +optional
+	Postgres *CommonSpec `json:"postgres,omitempty"`
+
+	// The spec of global hub manager
+	// +optional
+	Manager *CommonSpec `json:"manager,omitempty"`
+
+	// The spec of global hub agent
+	// +optional
+	Agent *CommonSpec `json:"agent,omitempty"`
+}
+
+type CommonSpec struct {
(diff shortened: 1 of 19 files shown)


operator/apis/v1alpha4/multiclusterglobalhub_types.go
```diff
@@ -90,6 +90,41 @@ type MulticlusterGlobalHubSpec struct {
 	// +kubebuilder:default={postgres: {retention: "18m"}}
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	DataLayer DataLayerConfig `json:"dataLayer"`
+	// Advanced configurations for global hub
+	// +optional
+	AdvancedConfig *AdvancedConfig `json:"advanced,omitempty"`
```
yanmxa, line 95: I suggest naming the definition more specific, for example, "ComponentResource" or even simpler like "Resource".

what do you think of it?

operator/apis/v1alpha4/multiclusterglobalhub_types.go
```diff
@@ -90,6 +90,41 @@ type MulticlusterGlobalHubSpec struct {
...
+	// The spec of global hub manager
+	// +optional
+	Manager *CommonSpec `json:"manager,omitempty"`
+
+	// The spec of global hub agent
+	// +optional
+	Agent *CommonSpec `json:"agent,omitempty"`
+}
+
+type CommonSpec struct {
```
yanmxa, line 124: The AdvancedConfig and CommonSpec mean there are other fields such as "BasicConfig" and "SpecialSpec". 

Is this your intention? I'm wondering if the customer can intuitively grasp the purpose of this API

operator/pkg/controllers/addon/addon_controller_manifests.go
```diff
@@ -66,6 +69,16 @@ type ManifestsConfig struct {
 	AgentQPS               float32
 	AgentBurst             int
 	LogLevel               string
+	// cannot use *corev1.ResourceRequirements, addonfactory.StructToValues removes the real value
+	Resources *Resources
+}
+
+type Resources struct {
```
yanmxa, line 76: Do we need to define the `Resources` type here? Or just leverage the [ResourceRequirements](https://github.com/kubernetes/api/blob/84c4768022421f4d420a8d74e6699077021563ce/core/v1/types.go#L2547) align with the manger and grafana.

operator/apis/v1alpha4/multiclusterglobalhub_types.go
clyang82: We may have more requirements in the future to configure for advanced usage. so we can put them into `advanced`.

operator/apis/v1alpha4/multiclusterglobalhub_types.go
clyang82: Yes. `CommonSpec` is a common one. In the future, we may define `AgentSpec` to include `CommonSpec` and other specific fields.

operator/pkg/controllers/addon/addon_controller_manifests.go
clyang82: check my comments for the `Resources` Type definition
> // cannot use *corev1.ResourceRequirements, addonfactory.StructToValues removes the real value

PR #670, opened 2023-10-10
Prune resources from database when detaching the hub from the global hub

yanmxa:
Signed-off-by: myan <myan@redhat.com>
Ref: https://issues.redhat.com/browse/ACM-7869, https://issues.redhat.com/browse/ACM-7873, https://issues.redhat.com/browse/ACM-7544

Note:
- Mark the data from `local_spec.policy`, `status.managed_clusters`, and `status.leaf_hub` as deleted status after detaching the hub.
- Delete the record from `local_status.compliance` when deleting the hub cluster
- Remove the deleted status when rejoining the hub to the global hub

PR #756, opened 2024-01-09
Remove the managed hub resources when its status isn't available

yanmxa:

resolved: https://issues.redhat.com/browse/ACM-8849 and https://issues.redhat.com/browse/ACM-9139

There are 2 ways to get the status of the managed hub:
- **ACM Cluster Lifecycle**
  - **Advantage**: Leverage the ACM cluster management from the global hub. when the cluster switches to unknown/detaching, we will clean up the database resource of this hub.
  - **Disadvantage**: to make sure the manager prevents receiving the message from the hub when it switches to unknown.
- **Heartbeat Lifecycle**: https://github.com/stolostron/multicluster-global-hub/pull/762
  - **Advantage**: Report the managed hub heartbeat by the agent, Define the `active`/`inactive` of the agent(hub) status by it. Clean the resources if the status switches to `inactive`. (Add a label to the ACM cluster)
  - **Disadvantage**: Need a mechanism(goroutine/controller) to check the timestamp of the heartbeat table and update the status of the hub
  
The latter will need us to find a new way the manage the lifecycle of the managed hub cluster. That would be a larger scope and the merit is it can be independent from the ACM framework. 

Note: We have used the ACM cluster for [the detaching scenario](https://github.com/stolostron/multicluster-global-hub/pull/670).


yanmxa:
- cluster  unknown(network issue):
  - hibernate
  - spoken cluster destroy
- the agent might run!

manager/pkg/controller/managed_hub_cluster.go
```diff
@@ -0,0 +1,146 @@
...
+	// the managed hub is deleted
+	if err != nil && errors.IsNotFound(err) {
+		return ctrl.Result{}, nil
+	} else if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// the managed hub is deleting then delete the data from database and then remove the finalizer
+	if !cluster.DeletionTimestamp.IsZero() || !clusterAvailable(cluster) {
+		logger.Info("remove database resources from the hub cluster", "hubAvailable", clusterAvailable(cluster))
```
clyang82, line 70: Please help me recall our discussion. if the cluster is not available, but the global agent is still working, do we need to remove the managed hub info from the DB?

yanmxa: We will add the heartbeat to determine the status of managed hub in this PR: https://github.com/stolostron/multicluster-global-hub/pull/762

PR #758, opened 2024-01-10
Enable metrics logs

yanmxa:

Ref: https://issues.redhat.com/browse/ACM-9125

### DIFF
--- a/manager/pkg/cronjob/task/local_compliance_history.go
+++ b/manager/pkg/cronjob/task/local_compliance_history.go
@@ -65,7 +65,7 @@ func SyncLocalCompliance(ctx context.Context, enableSimulation bool, job gocron.
 
 	historyDate := startTime.AddDate(0, 0, -interval)
 	log = ctrl.Log.WithName(LocalComplianceTaskName).WithValues("history", historyDate.Format(dateFormat))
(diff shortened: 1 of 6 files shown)


manager/pkg/cronjob/task/local_compliance_history.go
```diff
@@ -75,7 +75,7 @@ func SyncLocalCompliance(ctx context.Context, enableSimulation bool, job gocron.
 		log.Error(err, "sync from local_status.compliance to history.local_compliance failed")
 		return
 	}
-	log.V(2).Info("with local_status.compliance", "totalCount", statusTotal, "insertedCount", statusInsert)
+	log.Info("with local_status.compliance", "totalCount", statusTotal, "insertedCount", statusInsert)
```
clyang82, line 78: Can we pass `-v=2` to enable it?

yanmxa: But I mean to enable it at the information level, cause the cronjob only executes once a day, It won't produce much log but it's helpful for troubleshooting!

clyang82: but you can get such information in database, correct?

yanmxa: modify the log level to V(2)

PR #759, opened 2024-01-11
fix watch no spec resources

ldpliu:
The generation field is always 0 for no spec resources.
So we should update the watch logic for these kind of resources.

### DIFF
--- a/operator/pkg/controllers/hubofhubs/globalhub_controller.go
+++ b/operator/pkg/controllers/hubofhubs/globalhub_controller.go
@@ -296,7 +296,11 @@ var ownPred = predicate.Funcs{
 		return false
 	},
 	UpdateFunc: func(e event.UpdateEvent) bool {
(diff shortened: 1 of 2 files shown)


operator/pkg/controllers/hubofhubs/globalhub_controller.go
```diff
@@ -296,7 +296,11 @@ var ownPred = predicate.Funcs{
 		return false
 	},
 	UpdateFunc: func(e event.UpdateEvent) bool {
-		return e.ObjectOld.GetGeneration() != e.ObjectNew.GetGeneration() // only requeue when spec change
+		// only requeue when spec change, if the resource do not have spec field, the generation is always 0
+		if e.ObjectNew.GetGeneration() == 0 {
+			return true
+		}
+		return e.ObjectOld.GetGeneration() != e.ObjectNew.GetGeneration()
```
yanmxa, line 303: For the secret, this condition will always be false.

yanmxa:
Good Catch! Thanks @ldpliu! 

/lgtm

PR #767, opened 2024-01-14
watch kafka resources

clyang82:
Follow the following issues:
1. dynamically add watch kafka resources once the kafka CRD exists
2.  add tolerations and nodeselector to kafka resource. fixed: https://issues.redhat.com/browse/ACM-9245
3. fixed: https://issues.redhat.com/browse/ACM-9320


operator/pkg/controllers/hubofhubs/globalhub_middleware_controller.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (m *middlewareController) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
+	// get the mcgh cr name and then trigger the globalhub reconciler
+	mghList := &globalhubv1alpha4.MulticlusterGlobalHubList{}
+	err := m.mgr.GetClient().List(ctx, mghList)
+	if err != nil {
+		klog.Error(err, "Failed to list MulticlusterGlobalHub")
+		return ctrl.Result{}, err
+	}
+	_, err = m.reconciler.ReconcileTransport(ctx, &mghList.Items[0])
```
yanmxa, line 34: We can retrieve the mgh without list
```bash
	mgh := &globalhubv1alpha4.MulticlusterGlobalHub{}
	err := r.Get(ctx, config.GetMGHNamespacedName(), mgh)
```

operator/pkg/transporter/strimzi_transporter.go
```diff
@@ -576,20 +576,84 @@ func (k *strimziTransporter) kafkaClusterReady() error {
...
+	if updatedKafka.Spec.Kafka.Template.Pod.Affinity == nil {
+		updatedKafka.Spec.Kafka.Template.Pod.Affinity = desiredKafka.Spec.Kafka.Template.Pod.Affinity
+	}
+	updatedKafka.Spec.Kafka.Template.Pod.Affinity.NodeAffinity = desiredKafka.Spec.Kafka.Template.Pod.
+		Affinity.NodeAffinity
+	updatedKafka.Spec.Kafka.Template.Pod.Tolerations = desiredKafka.Spec.Kafka.Template.Pod.Tolerations
+	updatedKafka.Spec.Kafka.Resources = desiredKafka.Spec.Kafka.Resources
+
+	// update the zookeeper custom resource
+	if updatedKafka.Spec.Zookeeper.Template == nil {
```
yanmxa, line 609: Can we leverage the `DeepEqual` utils?

operator/main.go
```diff
@@ -236,7 +239,11 @@ func doMain(ctx context.Context, cfg *rest.Config) int {
 		return 1
 	}
 
-	// +kubebuilder:scaffold:builder
+	// start crd controller
+	if err = hubofhubscontrollers.StartCRDController(mgr, r); err != nil {
```
yanmxa, line 243: We need to detect the [TransportProtocol](https://github.com/stolostron/multicluster-global-hub/blob/7e34817f28cbc8ca8accfd8589dbadad7a8c55fc/pkg/transport/type.go#L66)(maybe we can put it to the config variables), only if the type is `StrimziTransporter`, then we add the crd controller to manager

operator/pkg/controllers/hubofhubs/globalhub_middleware_controller.go
yanmxa: Also, if we trigger to `ReconcileTransport ` here, maybe we need to remove the it from the `globalhub_middleware.go`

operator/pkg/transporter/strimzi_transporter.go
clyang82: we just focus on `Nodeaffinity` and `Tolerations`. User can add other attribute in `Template`. so we cannot leverage `DeepEqual` to compare directly.

operator/main.go
clyang82: Good point.

operator/pkg/controllers/hubofhubs/globalhub_middleware_controller.go
yanmxa: Global Controller -> ReconcileTransport -> propagation(credential manager/agent)

Kafka Controller -> ReconcileTransport(update/create credential)

PR #773, opened 2024-01-16
Update the managed hub status by the heartbeat

yanmxa:

Resolved: https://issues.redhat.com/browse/ACM-9313


manager/pkg/statussyncer/hubmanagement/hub_management.go
```diff
@@ -0,0 +1,85 @@
...
+		return err
+	}
+	if err := h.inactive(ctx, expiredHubs); err != nil {
+		return fmt.Errorf("inactive the hubs with error: %v", err)
+	}
+	return nil
+}
+
+func (h *hubManagement) inactive(ctx context.Context, hubs []models.LeafHubHeartbeat) error {
+	// TODO: cleanup the database resources
```
clyang82, line 73: Need to think how to handle reactive. maybe resync the status once the cluster is back after agent session timeout.

manager/cmd/manager/main.go
```diff
@@ -159,6 +160,8 @@ func parseFlags() *managerconfig.ManagerConfig {
 		"data retention indicates how many months the expired data will kept in the database")
 	pflag.BoolVar(&managerConfig.EnableGlobalResource, "enable-global-resource", false,
 		"enable the global resource feature.")
+	pflag.StringVar(&managerConfig.AgentSessionTimeout, "agent-session-timeout", "90s",
+		"The interval to detect/update the managed hub status")
```
clyang82, line 164: Maybe 5*AgentHeartbeatInterval as agent-session-timeout. so we do not need to expose agent-session-timeout as a param.

clyang82 (changes requested):

manager/pkg/statussyncer/hubmanagement/hub_management.go
yanmxa: Add the reactive function

manager/cmd/manager/main.go
yanmxa: Done

operator/pkg/controllers/hubofhubs/upgrade/1.upgrade.sql
```diff
@@ -4,3 +4,4 @@ ALTER TABLE status.leaf_hubs DROP CONSTRAINT leaf_hubs_pkey;
 ALTER TABLE status.leaf_hubs ADD CONSTRAINT leaf_hubs_pkey PRIMARY KEY (cluster_id, leaf_hub_name);
 
 ALTER TABLE status.leaf_hub_heartbeats ADD COLUMN IF NOT EXISTS status VARCHAR(10) DEFAULT 'active';
+CREATE INDEX IF NOT EXISTS leaf_hub_heartbeats_leaf_hub_status_idx ON status.leaf_hub_heartbeats(status);
```
clyang82, line 7: will it locate to `2.tables.sql` or `1.upgrade.sql`? since `2.tables.sql` will be executed always.

manager/pkg/statussyncer/hubmanagement/hub_management.go
```diff
@@ -0,0 +1,99 @@
...
+		Update("status", HubInactive).Error
+	if err != nil {
+		return err
+	}
+	if err := h.inactive(ctx, expiredHubs); err != nil {
+		return err
+	}
+
+	var reactiveHubs []models.LeafHubHeartbeat
+	err = db.Model(&reactiveHubs).
```
clyang82, line 75: What is your thoughts in reactiveFunction? if you delete the records from table in inactive function, you cannot get records for reactiving.

operator/pkg/controllers/hubofhubs/upgrade/1.upgrade.sql
yanmxa: I suggest putting the add(add table, index, and so on) SQL to the database folder. And put the modify action(add a column to a table, modify the primary key) to the update folder.

What do you think of it?

manager/pkg/statussyncer/hubmanagement/hub_management.go
yanmxa: The `reactiveFunction` will do the following actions:
1. Update the records of which hubs' status from inactive to active
2. Get these updating hubs and resync the bundles...

operator/pkg/controllers/hubofhubs/upgrade/1.upgrade.sql
clyang82: every change should be in `2.tables.sql` firstly. if need handle for upgrading, then put the logics in `1.upgrade.sql`. for example: You add a new column, it should be in both.

yanmxa: Done

operator/pkg/controllers/hubofhubs/database/2.tables.sql
```diff
@@ -28,7 +28,11 @@ CREATE TABLE IF NOT EXISTS status.leaf_hub_heartbeats (
     leaf_hub_name character varying(254) NOT NULL,
     last_timestamp timestamp without time zone DEFAULT now() NOT NULL
 );
+ALTER TABLE status.leaf_hub_heartbeats ADD COLUMN IF NOT EXISTS status VARCHAR(10) DEFAULT 'active';
```
clyang82, line 31: this one should be in `upgrade.sql` file.

yanmxa: It's already in there. https://github.com/stolostron/multicluster-global-hub/blob/2bf5ae936cff3c6c1a0e18e29c5a022b288d297b/operator/pkg/controllers/hubofhubs/upgrade/1.upgrade.sql#L7

PR #785, opened 2024-01-30
Cleanup the heartbeat records based on the retention

yanmxa:

### DIFF
--- a/manager/pkg/specsyncer/spec2db/controller/managedhub_spec_sync.go
+++ b/manager/pkg/specsyncer/spec2db/controller/managedhub_spec_sync.go
@@ -8,7 +8,6 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
-	"gorm.io/gorm"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


manager/pkg/specsyncer/spec2db/controller/managedhub_spec_sync.go
```diff
@@ -67,34 +64,6 @@ func (r *managedHubReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 
 	// the managed hub is deleting then delete the data from database and then remove the finalizer
 	if !cluster.DeletionTimestamp.IsZero() {
-		db := database.GetGorm()
```
clyang82, line 70: Do not need to softly delete those resources?

yanmxa: No. We delete these resources on the [hub_management.go](https://github.com/stolostron/multicluster-global-hub/blob/main/manager/pkg/statussyncer/hubmanagement/hub_management.go)

PR #786, opened 2024-01-30
do not backup kafka pvc

ldpliu:
Kafka auto add "cluster.open-cluster-management.io/backup=globalhub" label to entity operator and all kafka pvcs, so we need to exclude it.

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -32,8 +32,17 @@ const (
 
 	BackupKey             = "cluster.open-cluster-management.io/backup"
 	BackupVolumnKey       = "cluster.open-cluster-management.io/volsync"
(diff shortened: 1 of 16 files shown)


clyang82 (approved):
LGTM

pkg/constants/constants.go
```diff
@@ -32,8 +32,17 @@ const (
...
 	BackupActivationValue = "cluster-activation"
 	BackupGlobalHubValue  = "globalhub"
+
+	PostgresPvcLabelKey   = "component"
+	PostgresPvcLabelValue = "multicluster-global-hub-operator"
+)
+
+const (
+	//lock the database
+	LockId = "1"
```
yanmxa, line 45: Why add the LockId ?

PR #787, opened 2024-01-31
add lock when backup database

ldpliu:
story: https://issues.redhat.com/browse/ACM-9273
Use advisory lock to lock the db when backup postgres pvc.

### DIFF
--- a/pkg/conflator/conflation_committer.go
+++ b/pkg/conflator/conflation_committer.go
@@ -40,6 +40,7 @@ func (k *ConflationCommitter) Start(ctx context.Context) error {
 		ticker := time.NewTicker(time.Second * 5)
 
 		defer ticker.Stop()
+
 		for {
 			select {
 			case <-ticker.C: // wait for next time interval
@@ -88,6 +89,13 @@ func (k *ConflationCommitter) commit() error {
 	}
 
 	db := database.GetGorm()
+	conn := database.GetConn()
+
+	err := database.Lock(conn)
+	defer database.Unlock(conn)
+	if err != nil {
+		return err
+	}
 	if len(databaseTransports) > 0 {
 		err := db.Clauses(clause.OnConflict{
 			UpdateAll: true,
--- a/pkg/conflator/workerpool/worker.go
+++ b/pkg/conflator/workerpool/worker.go
@@ -7,6 +7,7 @@ import (
 	"github.com/go-logr/logr"
 
 	"github.com/stolostron/multicluster-global-hub/pkg/bundle"
+	"github.com/stolostron/multicluster-global-hub/pkg/database"
 	"github.com/stolostron/multicluster-global-hub/pkg/statistics"
 )
 
@@ -56,7 +57,17 @@ func (worker *Worker) start(ctx context.Context) {
 
 		case job := <-worker.jobsQueue: // DBWorker received a job request.
 			startTime := time.Now()
-			err := job.handlerFunc(ctx, job.bundle) // db connection released to pool when done
+			conn := database.GetConn()
+
+			err := database.Lock(conn)
+			if err != nil {
+				worker.log.Error(err, "failed to get db lock")
+				database.Unlock(conn)
+				continue
+			}
+			err = job.handlerFunc(ctx, job.bundle) // db connection released to pool when done
+			database.Unlock(conn)
+
 			worker.statistics.AddDatabaseMetrics(job.bundle, time.Since(startTime), err)
 			job.conflationUnitResultReporter.ReportResult(job.bundleMetadata, err)
 
--- a/pkg/database/gorm_conn.go
+++ b/pkg/database/gorm_conn.go
@@ -1,6 +1,7 @@
 package database
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
 	"net/url"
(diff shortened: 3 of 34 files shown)


pkg/database/gorm_conn.go
```diff
@@ -80,10 +88,39 @@ func GetGorm() *gorm.DB {
...
+}
+
+func Unlock(gormConn *gorm.DB) {
+	if !IsBackupEnabled {
+		return
+	}
+	log.V(2).Info("unlock db")
+	defer log.V(2).Info("db unlocked")
+	err := gormConn.Exec("select pg_advisory_unlock(?)", constants.LockId).Error
+	if err != nil {
```
clyang82, line 115: Need retry to ensure the lock is released.

manager/pkg/statussyncer/syncers/policy_compliance_handler.go
```diff
@@ -27,8 +27,14 @@ func (syncer *CompliancesDBSyncer) handleComplianceBundle(ctx context.Context,
 	leafHubName := bundle.GetLeafHubName()
 	db := database.GetGorm()
 
+	err := database.Lock(db)
+	if err != nil {
```
clyang82, line 31: do we have retry to write DB if there is an error?

pkg/database/gorm_conn.go
```diff
@@ -39,39 +43,43 @@ func InitGormInstance(config *DatabaseConfig) error {
...
+	})
+	if err != nil {
+		log.Error(err, "failed to open gorm connection")
+		return nil, nil, err
+	}
+	return gormDBconn, sqlDBConn, nil
+}
+
 func GetGorm() *gorm.DB {
 	if gormDB == nil {
```
clyang82, line 84: How about have 2 functions? one is `GetReadGorm()`, the other is `GetWriteGorm`. in `GetWriteGorm`, request the lock before return.

clyang82: do not need to request the lock for ReadGorm.

pkg/database/gorm_conn.go
```diff
@@ -80,10 +88,39 @@ func GetGorm() *gorm.DB {
...
+	return sqlDB
+}
+
+func Lock(gormConn *gorm.DB) error {
+	if !IsBackupEnabled {
+		return nil
+	}
+	log.V(2).Info("Add db lock")
+	defer log.V(2).Info("db locked")
+	return gormConn.Exec("select pg_advisory_lock(?)", constants.LockId).Error
```
clyang82, line 105: what is default timeout to wait for the lock?

pkg/database/gorm_conn.go
ldpliu: > Need retry to ensure the lock is released.

Done

manager/pkg/statussyncer/syncers/policy_compliance_handler.go
ldpliu: > do we have retry to write DB if there is an error?

I think we have retry logic when write db error. @yanmxa did that previously.

pkg/database/gorm_conn.go
ldpliu: We just only need to call db.lock() for write action.
For read action, there is no need to call db.lock()

pkg/database/gorm_conn.go
ldpliu: > what is default timeout to wait for the lock?

pg advisory lock look always hang if it do not get it. And the lock release when the session close.

manager/pkg/statussyncer/syncers/policy_compliance_handler.go
yanmxa: Yes, It will retry on the upper layer on the conflationReadyQueue.

pkg/database/gorm_conn.go
yanmxa: Can we just make the `GetGorm` a blocking call/invoke? 
We will return the instance until the following code unblocked
```go
gormConn.Exec("select pg_advisory_lock(?)", constants.LockId).Error
```

pkg/conflator/conflation_committer.go
```diff
@@ -40,6 +40,8 @@ func (k *ConflationCommitter) Start(ctx context.Context) error {
 		ticker := time.NewTicker(time.Second * 5)
 
 		defer ticker.Stop()
+		defer database.Unlock(database.GetGorm())
```
clyang82, line 43: may not require unlock here?

pkg/database/gorm_conn.go
```diff
@@ -80,10 +90,48 @@ func GetGorm() *gorm.DB {
...
+func GetSqlDb() *sql.DB {
+	if sqlDB == nil {
+		log.Error(nil, "sqlDb connection is not initialized")
+		return nil
+	}
+	return sqlDB
+}
+
+func Lock(gormConn *gorm.DB) error {
+	if !IsBackupEnabled {
```
clyang82, line 102: Can you add tests for lock and unlock? for example: cannot get lock twice in the same time / if the lock is not released, wait for until timeout to unlock automatically.

pkg/conflator/workerpool/worker.go
```diff
@@ -56,7 +57,15 @@ func (worker *Worker) start(ctx context.Context) {
 
 		case job := <-worker.jobsQueue: // DBWorker received a job request.
 			startTime := time.Now()
-			err := job.handlerFunc(ctx, job.bundle) // db connection released to pool when done
+			db := database.GetGorm()
+			err := database.Lock(db)
+			if err != nil {
+				worker.log.Error(err, "failed to get db lock")
+				continue
```
yanmxa, line 64: If it continues to the next loop, it may cause deadlock!

pkg/conflator/conflation_committer.go
ldpliu: done

pkg/conflator/workerpool/worker.go
ldpliu: done

pkg/database/gorm_conn.go
ldpliu: I have add a test for lock, but it's difficult to simulate unlock fail, and the unlock will not do auto unlock, we just retry when lock fail. it will unlock when session exit.

morvencao (commented):
/approve
/lgtm

PR #789, opened 2024-02-01
Amq dashboards

clyang82:
ref: https://issues.redhat.com/browse/ACM-9754


operator/apis/v1alpha4/multiclusterglobalhub_types.go
```diff
@@ -93,6 +93,9 @@ type MulticlusterGlobalHubSpec struct {
 	// Advanced configurations for global hub
 	// +optional
 	AdvancedConfig *AdvancedConfig `json:"advanced,omitempty"`
+	// EnableMetrics enables the metrics for the global hub kafka components
+	// +optional
+	EnableMetrics bool `json:"enableMetrics,omitempty"`
```
yanmxa, line 98: Is the default value `false`?

operator/pkg/controllers/hubofhubs/manifests/kafka/podmonitor.yaml
```diff
@@ -0,0 +1,66 @@
+apiVersion: monitoring.coreos.com/v1
```
yanmxa, line 1: Suggest adding a prefix(`kafka-metrics`) to these two pod monitors.

And I have a question: Can we combine these two monitors to one?

operator/pkg/controllers/hubofhubs/manifests/kafka/metrics.yaml
```diff
@@ -0,0 +1,184 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: kafka-metrics
```
yanmxa, line 4: Same as above

operator/pkg/controllers/hubofhubs/manifests/grafana/serviceaccount-secret.yaml
```diff
@@ -0,0 +1,10 @@
+{{- if .EnableMetrics }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: multicluster-global-hub-grafana-sa-secret
```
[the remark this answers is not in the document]
clyang82: Do you mean as the prefix for the name? if so, I do not think so. because the secret is bundled to sa, it can be used for other purpose.

operator/apis/v1alpha4/multiclusterglobalhub_types.go
yanmxa: If it is just a feature for dev or internal, putting it into the API will expose it to customers? maybe we should put some where the customer can't sense it.

PR #831, opened 2024-03-01
Update to use golang 1.21

clyang82:
Each major Go release is supported until there are two newer major releases. For example, Go 1.5 was supported until the Go 1.7 release, and Go 1.6 was supported until the Go 1.8 release. We fix critical problems, including critical security problems, in supported releases as needed by issuing minor revisions.

1.22 was just released on 2/6/24, so 1.20 is already end-of-life


clyang82:
release-2.10 was updated to 1.21 in the past.

PR #837, opened 2024-03-05
Fix the bug that local compliance history data conflicts

yanmxa:
Signed-off-by: myan <myan@redhat.com>

Error message:
```bash
2024/03/05 00:29:55 /remote-source/multicluster-global-hub/app/manager/pkg/cronjob/task/local_compliance_history.go:282 pq: ON CONFLICT DO UPDATE command cannot affect row a second time
[3.439ms] [rows:0]
			INSERT INTO history.local_compliance (policy_id, cluster_id, leaf_hub_name, compliance_date, compliance,
					compliance_changed_frequency)
			WITH compliance_aggregate AS (
					SELECT cluster_id, policy_id, leaf_hub_name,
							CASE
									WHEN bool_or(compliance = 'non_compliant') THEN 'non_compliant'
									WHEN bool_or(compliance = 'unknown') THEN 'unknown'
									ELSE 'compliant'
							END::local_status.compliance_type AS aggregated_compliance
					FROM event.local_policies
					WHERE created_at BETWEEN CURRENT_DATE - INTERVAL '1 days' AND CURRENT_DATE - INTERVAL '0 day'
					GROUP BY cluster_id, policy_id, leaf_hub_name
			)
			SELECT policy_id, cluster_id, leaf_hub_name, (CURRENT_DATE - INTERVAL '1 day'), aggregated_compliance,
					(SELECT COUNT(1) FROM (
							SELECT created_at, compliance,
									LAG(compliance) OVER (PARTITION BY cluster_id, policy_id ORDER BY created_at ASC)
									AS prev_compliance
							FROM event.local_policies lp
							WHERE (lp.created_at BETWEEN CURRENT_DATE - INTERVAL '1 days' AND CURRENT_DATE - INTERVAL '0 day')
									AND lp.cluster_id = ca.cluster_id AND lp.policy_id = ca.policy_id
							ORDER BY created_at ASC
					) AS subquery WHERE compliance <> prev_compliance) AS compliance_changed_frequency
			FROM compliance_aggregate ca
			ORDER BY cluster_id, policy_id
			LIMIT 1000 OFFSET 0
			ON CONFLICT (policy_id, cluster_id, compliance_date)
			DO UPDATE SET
				compliance = EXCLUDED.compliance,
				compliance_changed_frequency = EXCLUDED.compliance_changed_frequency;

[...]

PR #838, opened 2024-03-05
Fixed the history compliance conflicts with exist condition

yanmxa:
Reference: https://issues.redhat.com/browse/ACM-10255

It also needs to be noted this change will lose the history data in the following cases:

If a managed hub cluster was detached yesterday, then the history compliance of yesterday will not persist on the database.


clyang82:
The fix will lead to lose data if the managed hub cluster is detached. the historical data for that day which the managed hub is detached will be missed.

clyang82:
We have doc-ed a known issue in this release. close this PR and in favor of https://github.com/stolostron/multicluster-global-hub/pull/837

PR #850, opened 2024-03-20
Reduce the manager producer footprint

yanmxa:
Reference: https://issues.redhat.com/browse/ACM-10642

image

### DIFF
--- a/pkg/transport/config/confluent_config.go
+++ b/pkg/transport/config/confluent_config.go
@@ -22,6 +22,8 @@ func GetConfluentConfigMap(kafkaConfig *transport.KafkaConfig, producer bool) (*
 		"ssl.endpoint.identification.algorithm": "none",
 	}
 	if producer {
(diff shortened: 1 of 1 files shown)


pkg/transport/config/confluent_config.go
```diff
@@ -22,6 +22,8 @@ func GetConfluentConfigMap(kafkaConfig *transport.KafkaConfig, producer bool) (*
 		"ssl.endpoint.identification.algorithm": "none",
 	}
 	if producer {
+		_ = kafkaConfigMap.SetKey("go.events.channel.size", 1000)
```
clyang82, line 25: What is the default value?

clyang82 (approved):
LGTM

pkg/transport/config/confluent_config.go
yanmxa: The default is `1000000`

PR #853, opened 2024-03-25
Fix global hub inspector chart issue

yanmxa:

If the value isn't ordered on the x-axis, then the chart is not plotted in chronological order.


clyang82 (approved):
LGTM

PR #861, opened 2024-03-29
Sync policy object only policy spec is changed

yanmxa:

Reference: https://issues.redhat.com/browse/ACM-10797

For the policy spec event, we only need to report the object when the generation(spec) is changed. And ignore the status updates.

### DIFF
--- a/agent/pkg/status/controller/apps/subscription_report_sync.go
+++ b/agent/pkg/status/controller/apps/subscription_report_sync.go
@@ -23,7 +23,7 @@ func LaunchSubscriptionReportSyncer(ctx context.Context, mgr ctrl.Manager, agent
 	predicate := predicate.NewPredicateFuncs(func(object client.Object) bool { return true })
 
 	// emitter config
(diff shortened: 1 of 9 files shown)


clyang82:
Could you update the description why we need this changes? Is it introduced by the code refactor?

agent/pkg/status/controller/apps/subscription_report_sync.go
```diff
@@ -23,7 +23,7 @@ func LaunchSubscriptionReportSyncer(ctx context.Context, mgr ctrl.Manager, agent
 	predicate := predicate.NewPredicateFuncs(func(object client.Object) bool { return true })
 
 	// emitter config
-	emitter := generic.ObjectEmitterWrapper(enum.SubscriptionReportType, nil, nil)
+	emitter := generic.ObjectEmitterWrapper(enum.SubscriptionReportType, nil, nil, false)
```
clyang82, line 26: add TODO to handle it if needed.

yanmxa: Done.

clyang82 (approved):
LGTM

PR #862, opened 2024-03-29
Fix the bug that manager might cleanup compliance after it restart

yanmxa:

Reference: https://issues.redhat.com/browse/ACM-10798


clyang82 (approved):
LGTM

PR #864, opened 2024-03-29
Enable the manager conflator to process the delta bundle 

yanmxa:

1. The conflation model handles full-state bundles like `compliance`, `managed clusters`... However, it might lose some data with the delta bundle, like `policy events`. 

2. This PR aims to enable the conflation support of both `complete` and `delta` bundles without losing data.

3. This Bug doesn't appear in Global Hub 1.1.0, Cause we use a goroutine to handle the event bundle.

4. Using the buffer channel to replace the [Mutex List](https://github.com/stolostron/multicluster-global-hub/pull/864/files#diff-42be6b645d39b8f8f44f594ee6ec1546d31c7abdd3b75659142bf55cd51c922cL24-L58) also reduce the manager's footprint significantly!
image

Reference: https://issues.redhat.com/browse/ACM-10804
Block by this PR: https://github.com/stolostron/multicluster-global-hub/pull/862. The current PR contain it's updates


yanmxa:
Now we can see with 75000 compliance and it also reports 75000 events to the database:
image
image

manager/pkg/statussyncer/conflator/conflation_unit.go
```diff
@@ -33,11 +34,20 @@ type ConflationUnit struct {
...
+		// requireInitialDependencyChecks: requireInitialDependencyChecks,
+		isInReadyQueue: false,
+		lock:           sync.Mutex{},
+		statistics:     statistics,
+	}
 
 	for _, registration := range registrations {
-		priorityQueue[registration.priority] = &conflationElement{
+		conflationUnit.ElementPriorityQueue[registration.priority] = &conflationElement{
+			conflationUnit:       conflationUnit,
```
clyang82, line 50: why  `conflationElement` contains `conflationUnit` and `conflationUnit` contains `[]*conflationElement`? it does not make sense for me.

yanmxa:
/retest

I will refactor the `hubofhub` unit test the next day.

manager/pkg/statussyncer/conflator/conflation_unit.go
yanmxa: Done

manager/pkg/statussyncer/conflator/element_delta.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// Update is to update element payload
+func (e *deltaElement) Update(event *cloudevents.Event, metadata ConflationMetadata) {
+	e.job = NewConflationJob(event, metadata, e.handlerFunction, nil)
+}
+
+// the delta element will be delivered to job chan directly
+func (e *deltaElement) IsReadyToProcess(cu *ConflationUnit) bool {
+	return false
```
clyang82, line 72: always return false?

yanmxa: This method is only for the complete element. Remove it from the interface and now delta element doesn't need to care about it.

clyang82 (approved):
LGTM

PR #876, opened 2024-04-03
Enable placement syncers for global resouces

yanmxa:
## Summary

We will only enable the placement/placementrule/placementdecision for global resources and disable them for local resource

## Related issue(s)

Fixes #


yanmxa:
> why need this change? is it introduced by code refactor?

I just find that the placement, placementrule and placementdecision in the message queue. But we don't need it for now.

clyang82 (approved):
LGTM

PR #880, opened 2024-04-03
update olm descriptor reference

clyang82:
## Summary

refer to https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md#17-select update olm descriptor reference

the result is
image

## Related issue(s)

Fixes #


clyang82:
FYI @hchenxa

PR #886, opened 2024-04-10
:bug: fix kafka cluster error when enable metric by default

ldpliu:
## Summary
Globalhub can not install successfully when enable metric in mgh.
## Related issue(s)

Fixes #
When enable metric in mgh, kafka cluster depend on a kafka-metric configmap, but it's created after kafka operator and it depend on kafka cluster ready in current code logic, so we should create the configmap first.

### DIFF
--- a/operator/pkg/controllers/hubofhubs/globalhub_middleware.go
+++ b/operator/pkg/controllers/hubofhubs/globalhub_middleware.go
@@ -27,12 +27,18 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/discovery/cached/memory"
+	"k8s.io/client-go/restmapper"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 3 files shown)


operator/pkg/controllers/hubofhubs/globalhub_middleware.go
```diff
@@ -72,17 +72,17 @@ func (r *MulticlusterGlobalHubReconciler) ReconcileMiddleware(ctx context.Contex
 		// strimzi transporter -> use the kafka reconiler to get the connection
 		// wait until the kafka crd is ready, then start the kafka reconciler
 		if !r.KafkaInit {
+			err = addKafkaCRDController(r.Manager, r)
+			if err != nil {
+				errorChan <- err
+				return
+			}
```
clyang82, line 79: for the initialize case, the kafka crd does not exist, so that the KakfaCRDController starts to run, but it will not start KafkaController. it relies on `ReconcileTransport` to start Kafka. I do not know why it will fix the issue your described.

ldpliu: Changed

clyang82 (approved):

clyang82:
/override ci/prow/sonarcloud 
since it is a blocking issue, let me have a following PR to test this method @ldpliu

PR #897, opened 2024-04-24
:bug: Fix the topic isn't ready for the detach case

yanmxa:
## Summary

When detaching a managed hub cluster or disable the agent, should delete the shared topic like `event` or `spec`.

## Related issue(s)

Fixes # https://issues.redhat.com/browse/ACM-11284


clyang82 (approved):
LGTM

PR #898, opened 2024-04-24
:bug: add entity-operator monitor

ldpliu:
## Summary
The kafkauser and kafkatopic metric collected in entyty-operator, and we do not monitor it previously.
图片


clyang82 (approved):
LGTM

PR #899, opened 2024-04-24
:seedling: Persistent the agent event state to the configmap

yanmxa:
## Summary

Add the configMap to save the agent event state, so as to deduplicate the event in the message queue(event topics)

## Related issue(s)

Fixes #https://issues.redhat.com/browse/ACM-11243

### DIFF
--- a/agent/pkg/status/controller/event/local_replicated_policy_emitter.go
+++ b/agent/pkg/status/controller/event/local_replicated_policy_emitter.go
@@ -3,15 +3,16 @@ package event
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	cloudevents "github.com/cloudevents/sdk-go/v2"
 	"github.com/go-logr/logr"
-	lru "github.com/hashicorp/golang-lru"
 	corev1 "k8s.io/api/core/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/stolostron/multicluster-global-hub/agent/pkg/status/controller/config"
+	"github.com/stolostron/multicluster-global-hub/agent/pkg/status/controller/filter"
 	"github.com/stolostron/multicluster-global-hub/agent/pkg/status/controller/generic"
 	"github.com/stolostron/multicluster-global-hub/agent/pkg/status/controller/policies"
 	"github.com/stolostron/multicluster-global-hub/pkg/bundle/event"
@@ -60,29 +61,30 @@ var _ generic.ObjectEmitter = &localReplicatedPolicyEmitter{}
 
 type localReplicatedPolicyEmitter struct {
 	ctx             context.Context
+	name            string
 	log             logr.Logger
 	eventType       string
 	runtimeClient   client.Client
 	currentVersion  *version.Version
 	lastSentVersion version.Version
 	payload         event.ReplicatedPolicyEventBundle
-	cache           *lru.Cache
 	topic           string
 }
 
 func NewLocalReplicatedPolicyEmitter(ctx context.Context, runtimeClient client.Client,
 	topic string,
 ) generic.ObjectEmitter {
-	cache, _ := lru.New(20)
+	name := strings.Replace(string(enum.LocalReplicatedPolicyEventType), enum.EventTypePrefix, "", -1)
+	filter.RegisterTimeFilter(name)
 	return &localReplicatedPolicyEmitter{
 		ctx:             ctx,
-		log:             ctrl.Log.WithName("policy-event-syncer/replicated-policy"),
+		name:            name,
+		log:             ctrl.Log.WithName(name),
 		eventType:       string(enum.LocalReplicatedPolicyEventType),
 		topic:           topic,
 		runtimeClient:   runtimeClient,
 		currentVersion:  version.NewVersion(),
 		lastSentVersion: *version.NewVersion(),
-		cache:           cache,
 		payload:         make([]event.ReplicatedPolicyEvent, 0),
 	}
 }
@@ -115,9 +117,8 @@ func (h *localReplicatedPolicyEmitter) Update(obj client.Object) bool {
 	if !ok {
 		return false
 	}
-	// if exist, then return
-	evtKey := getEventKey(evt)
-	if h.cache.Contains(evtKey) {
+	// if it's a older event, then return false
+	if !filter.Newer(h.name, evt.LastTimestamp.Time) {
 		return false
 	}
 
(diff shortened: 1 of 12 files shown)


agent/pkg/status/controller/event/local_replicated_policy_emitter.go
```diff
@@ -143,15 +140,14 @@ func (h *localReplicatedPolicyEmitter) Update(obj client.Object) bool {
 			Reason:         evt.Reason,
 			Count:          evt.Count,
 			Source:         evt.Source,
-			CreatedAt:      evt.CreationTimestamp,
+			CreatedAt:      evt.LastTimestamp,
```
clyang82, line 143: what is the difference for those two timestamp?

agent/pkg/status/controller/filter/time_filter.go
```diff
@@ -0,0 +1,144 @@
...
+func LaunchTimeFilter(ctx context.Context, c client.Client, namespace string) error {
+	agentStateConfigMap := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      CACHE_CONFIG_NAME,
+			Namespace: namespace,
+		},
+	}
+	err := c.Get(ctx, client.ObjectKeyFromObject(agentStateConfigMap), agentStateConfigMap)
+	if err != nil && errors.IsNotFound(err) {
+		if e := c.Create(ctx, agentStateConfigMap); e != nil {
```
clyang82, line 53: do we need to add create permission?

agent/pkg/status/controller/filter/time_filter.go
```diff
@@ -0,0 +1,144 @@
...
+	return nil
+}
+
+// RegisterTimeFilter call before the LaunchTimeFilter, it will get the init time from the configMap
+func RegisterTimeFilter(key string) {
+	eventTimeCache[key] = time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC)
+	lastEventTimeCache[key] = time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC)
+}
+
+func loadTimeCacheFromConfigMap(cm *corev1.ConfigMap, key string) error {
```
clyang82, line 131: maybe `loadEventTimeCacheFromConfigmap`

agent/pkg/status/controller/filter/time_filter.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+		err = c.Update(ctx, cm, &client.UpdateOptions{})
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// RegisterTimeFilter call before the LaunchTimeFilter, it will get the init time from the configMap
```
clyang82, line 125: i saw you just initiate the map.  Do you need to get from the configmap? i think we need

clyang82 (changes requested):

agent/pkg/status/controller/event/local_replicated_policy_emitter.go
yanmxa: `CreationTimestamp` is from [ObjectMeta](https://kubernetes.io/zh-cn/docs/reference/kubernetes-api/common-definitions/object-meta/). It indicates [when the object is created](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#metadata).

`LastTimestamp` is the [time](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L6646) at which the most recent occurrence of this event was recorded. 

From the above description, we can see that the event can only be created once, meaning it cannot be changed even if it is updated. So instead of using `CreationTimestamp`, I choose the `LastTimestamp` as the unique event identifier for the time filter.

agent/pkg/status/controller/filter/time_filter.go
yanmxa: We already have the create permission. 
https://github.com/stolostron/multicluster-global-hub/blob/main/operator/pkg/controllers/addon/manifests/templates/agent/multicluster-global-hub-agent-clusterrole.yaml#L97-L101

agent/pkg/status/controller/filter/time_filter.go
yanmxa: It inits the map with an empty collection and restores value from the confgmap. 

I'm okay to init and restore value on `LaunchTimeFilter` if you think it is necessary.

agent/pkg/status/controller/filter/time_filter.go
yanmxa: Done.

agent/pkg/status/controller/event/local_replicated_policy_emitter.go
clyang82: I think we should keep using `CreationTimestamp`. To my understand, the `CreationTimestamp` is for real creation for this event. sometimes, the `firstTimestamp` and `lastTimestamp` are not set. Again, if the events have emitted in multiple times, the `firstTimestamp` and `lastTimestamp` are meaningful.
```
- action: DecisionUpdated
  apiVersion: v1
  eventTime: "2024-05-08T06:44:04.303377Z"
  firstTimestamp: null
  involvedObject:
    apiVersion: cluster.open-cluster-management.io/v1beta1
    kind: Placement
    name: global
    namespace: open-cluster-management-global-set
    resourceVersion: "149108"
    uid: 92f660b0-5383-43eb-a10d-429e946cdc5c
  kind: Event
  lastTimestamp: null
  message: Decision global-decision-1 is updated with placement global in namespace
    open-cluster-management-global-set
  metadata:
    creationTimestamp: "2024-05-08T06:44:04Z"
    name: global.17cd706c79fc438f
    namespace: open-cluster-management-global-set
    resourceVersion: "12338614"
    uid: 8fd3c9d7-2771-440f-bf70-0d6df20912b5
  reason: DecisionUpdate
  related:
    apiVersion: cluster.open-cluster-management.io/v1beta1
    kind: PlacementDecision
    name: global-decision-1
    namespace: open-cluster-management-global-set
    resourceVersion: "149105"
    uid: ceddbd89-d04a-4203-b006-94756be2e3bd
  reportingComponent: placementController
  reportingInstance: placementController-cluster-manager-placement-controller-6c8b7cf458-5xb8s
  series:
    count: 6
    lastObservedTime: "2024-05-08T06:44:12.674112Z"
  source: {}
  type: Normal
```

yanmxa: Yes. The other timestamp might be empty. Use the `creationTimestamp` back

clyang82 (approved):
LGTM

PR #905, opened 2024-04-30
update topic replica to 3

ldpliu:
## Summary
As we discussed in meeting, we should update the topic replica to 3.
## Related issue(s)

Fixes #

### DIFF
--- a/operator/pkg/transporter/strimzi_transporter.go
+++ b/operator/pkg/transporter/strimzi_transporter.go
@@ -66,7 +66,7 @@ var (
 	KafkaStorageDeleteClaim        = false
 	KafkaVersion                   = "3.6.0"
 	DefaultPartition         int32 = 1
-	DefaultPartitionReplicas int32 = 2
+	DefaultPartitionReplicas int32 = 3
 	// kafka metrics constants
 	KakfaMetricsConfigmapName       = "kafka-metrics"
 	KafkaMetricsConfigmapKeyRef     = "kafka-metrics-config.yml"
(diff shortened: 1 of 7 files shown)


operator/pkg/transporter/strimzi_transporter.go
```diff
@@ -261,9 +262,22 @@ func (k *strimziTransporter) CreateTopic(topic *transport.ClusterTopic) error {
...
+			if errors.IsNotFound(err) {
+				if e := k.runtimeClient.Create(k.ctx, k.newKafkaTopic(topicName)); e != nil {
+					return e
+				}
+				return nil
+			}
+			return err
+		}
+		desiredTopic := k.newKafkaTopic(topicName)
+		if !reflect.DeepEqual(desiredTopic.Spec.Replicas, kafkaTopic.Spec.Replicas) {
```
yanmxa, line 275: If we only assert the replicas, maybe we don't need the `reflect.DeepEqual` method

ldpliu: > If we only assert the replicas, maybe we don't need the `reflect.DeepEqual` method

Changed.

clyang82 (changes requested):
Need to update the document for the storage size

operator/pkg/transporter/strimzi_transporter.go
```diff
@@ -261,9 +261,22 @@ func (k *strimziTransporter) CreateTopic(topic *transport.ClusterTopic) error {
...
+		if err != nil {
+			if errors.IsNotFound(err) {
+				if e := k.runtimeClient.Create(k.ctx, k.newKafkaTopic(topicName)); e != nil {
+					return e
+				}
+				return nil
+			}
+			return err
+		}
+		desiredTopic := k.newKafkaTopic(topicName)
```
clyang82, line 273: after inner discussion, we agree to reconcile the kafkatopic and kafkauser. in other words, we do not allow the user to modify the kafkatopic and kafkauser. @hchenxa @yahliu any concerns for this behaviour change?

clyang82 (approved):
LGTM

PR #924, opened 2024-05-15
:bug: clean up metric resource when disable metrics

ldpliu:
## Related issue(s)

Fixes #

### DIFF
--- a/operator/pkg/controllers/hubofhubs/globalhub_controller.go
+++ b/operator/pkg/controllers/hubofhubs/globalhub_controller.go
@@ -201,27 +201,17 @@ func (r *MulticlusterGlobalHubReconciler) Reconcile(ctx context.Context, req ctr
 		return ctrl.Result{Requeue: true}, err
 	}
 
-	// Set BYO kafka and BYO postgres
-	err = config.SetBYOKafka(ctx, r.Client, mgh.Namespace)
-	if err != nil {
-		return ctrl.Result{Requeue: true}, err
-	}
-	err = config.SetBYOPostgres(ctx, r.Client, mgh.Namespace)
-	if err != nil {
-		return ctrl.Result{Requeue: true}, err
-	}
-
-	if config.IsBYOKafka() && config.IsBYOPostgres() {
-		mgh.Spec.EnableMetrics = false
-		klog.Info("Kafka and Postgres are provided by customer, disable metrics")
-	}
-
 	// only reconcile once: upgrade
 	r.upgradeOnce.Do(func() {
 		err = r.Upgrade(ctx)
 	})
 	if err != nil {
 		r.Log.Error(err, "failed to upgrade from release-2.10")
+		return ctrl.Result{}, err
+	}
+
+	err = r.reconcileMiddlewareMetrics(ctx, mgh)
+	if err != nil {
 		return ctrl.Result{}, err
 	}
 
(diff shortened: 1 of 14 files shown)


operator/pkg/controllers/hubofhubs/manifests/grafana/acm-global-postgres-exporter.yaml
```diff
@@ -3280,4 +3280,6 @@ kind: ConfigMap
 metadata:
   name: grafana-dashboard-acm-global-postgres-exporter
   namespace: {{ .Namespace }}
+  labels:
+    dependon: enableMetrics
```
yanmxa, line 3284: Can we align the owner label with other components key: `global-hub.open-cluster-management.io/managed-by`. And then just define the value like `operand-metrics`, `global-hub-metrics`?

ldpliu: > Can we align the owner label with other components key: `global-hub.open-cluster-management.io/managed-by`. And then just define the value like `operand-metrics`, `global-hub-metrics`?

As they already have the `global-hub.open-cluster-management.io/managed-by:global-hub-operator` label and it's meaningful. So I think use a new label to identity the metrics  is better. What do you think?

yanmxa: I mean the key `global-hub.open-cluster-management.io/managed-by` is used for owner relationships. so we can just use the key and give another value name for these metrics resources?

operator/pkg/controllers/hubofhubs/globalhub_controller.go
```diff
@@ -216,6 +216,13 @@ func (r *MulticlusterGlobalHubReconciler) Reconcile(ctx context.Context, req ctr
 		klog.Info("Kafka and Postgres are provided by customer, disable metrics")
 	}
 
+	if !mgh.Spec.EnableMetrics {
```
clyang82, line 219: maybe we define a new method `reconcileMetrics` to include all related with enableMetrics.

pkg/constants/constants.go
```diff
@@ -79,7 +79,9 @@ const (
 	// Deprecated identify the resource is a local-resource
 	// GlobalHubLocalResource = "global-hub.open-cluster-management.io/local-resource"
 	// if the resource with this label, it will be synced to database and then propagated to managed hub
-	GlobalHubGlobalResourceLabel = "global-hub.open-cluster-management.io/global-resource"
+	GlobalHubGlobalResourceLabel   = "global-hub.open-cluster-management.io/global-resource"
+	GlobalhubDependOnLabelKey      = "dependon"
```
clyang82, line 83: the label name should have this prefix `global-hub.open-cluster-management.io/`. so you can define the label name as `global-hub.open-cluster-management.io/metrics-resource`

BTW, you need to update this doc to reflect your changes - https://github.com/stolostron/multicluster-global-hub/tree/main/doc/reserved_names

ldpliu: Done

operator/pkg/controllers/hubofhubs/globalhub_controller.go
ldpliu: Done

clyang82 (approved):
LGTM

PR #925, opened 2024-05-15
:bug: Update the agent cache filter for cluster event

yanmxa:
## Summary

Fix the bug that event is missing from the database.

## Related issue(s)

Fixes #


clyang82 (approved):
LGTM

this is a blocker so merge it firstly. But we need to find a better solution to limit the cache since we only care about the policy events and managed cluster events.

PR #928, opened 2024-05-16
:seedling: Ignore the all broker is down error for producer client

yanmxa:
## Related issue(s)

Fixes # https://github.com/open-cluster-management-io/sdk-go/pull/44

### DIFF
--- a/pkg/transport/producer/generic_producer.go
+++ b/pkg/transport/producer/generic_producer.go
@@ -174,7 +174,15 @@ func handleProducerEvents(log logr.Logger, eventChan chan kafka.Event) {
 				// as the underlying client will automatically try to
 				// recover from any errors encountered, the application
 				// does not need to take action on them.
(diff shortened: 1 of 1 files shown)


pkg/transport/producer/generic_producer.go
```diff
@@ -174,7 +174,15 @@ func handleProducerEvents(log logr.Logger, eventChan chan kafka.Event) {
 				// as the underlying client will automatically try to
 				// recover from any errors encountered, the application
 				// does not need to take action on them.
-				log.Info("Transport producer client error, ignore it for most cases", "error", ev)
+				if ev.Code() == kafka.ErrAllBrokersDown {
```
clyang82, line 177: is it possible to use the sdk-go as a library in global hub?

yanmxa: It's refactoring all the global hub components. And we need to handle a lot of cases for the global hub

clyang82 (approved):

PR #933, opened 2024-05-21
:seedling: Add `cluster_name` and `cluster_id` for the managed cluster events

yanmxa:
## Summary

Note:
1. add cluster_name in the cluster event table
2. patch the cluster_id in the event table when managed_clusters is created/updated

## Related issue(s)

Fixes #

### DIFF
--- a/agent/pkg/status/controller/event/managedcluster_emitter.go
+++ b/agent/pkg/status/controller/event/managedcluster_emitter.go
@@ -94,6 +94,7 @@ func (h *managedClusterEmitter) Update(obj client.Object) bool {
 		EventNamespace: evt.Namespace,
 		Message:        evt.Message,
 		Reason:         evt.Reason,
+		ClusterName:    cluster.Name,
 		// ClusterID:           clusterId,
 		LeafHubName:         config.GetLeafHubName(),
 		ReportingController: evt.ReportingController,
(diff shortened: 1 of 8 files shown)


agent/pkg/status/controller/event/managedcluster_emitter.go
```diff
@@ -88,25 +89,47 @@ func (h *managedClusterEmitter) Update(obj client.Object) bool {
...
-		EventName:           evt.Name,
-		EventNamespace:      evt.Namespace,
-		Message:             evt.Message,
-		Reason:              evt.Reason,
-		ClusterID:           clusterId,
+		EventName:      evt.Name,
+		EventNamespace: evt.Namespace,
+		Message:        evt.Message,
+		Reason:         evt.Reason,
+		ClusterName:    cluster.Name,
```
clyang82, line 97: Checked with the original event, the involvedobject should include the managedcluster name. Can we have involvedObject here?

yanmxa: As we discussed, we will ignore the involveobject for now

operator/pkg/controllers/hubofhubs/database/3.functions.sql
```diff
@@ -20,6 +20,20 @@ BEGIN
...
+CREATE OR REPLACE FUNCTION public.update_cluster_event_cluster_id()
+    RETURNS TRIGGER
+    LANGUAGE plpgsql
+AS $$
+BEGIN
+    UPDATE event.managed_clusters
+    SET cluster_id = NEW.cluster_id
+    WHERE leaf_hub_name = NEW.leaf_hub_name
+        AND cluster_name = (NEW.payload -> 'metadata' ->> 'name')
+        AND cluster_id = (New.payload -> 'metadata' ->> 'uid')::uuid;
```
clyang82, line 32: is cluster_id empty here?

yanmxa: No, in the previous implementation. I get the cliamId, if it isn't exist, then use the kube uid.

The good news is the clusterId will always have values and even with the same cluster name we can distinguish their event. But the clusterId in the cluster event table might be changed to claimId once the instance in status.managed_cluster is created or updated.



also, the clusterId with uuid type seems cannot be empty string ""

pkg/database/models/event.go
```diff
@@ -52,13 +52,14 @@ func (DataRetentionJobLog) TableName() string {
 type ManagedClusterEvent struct {
 	EventNamespace      string    `gorm:"column:event_namespace;type:varchar(63);not null" json:"eventNamespace"`
 	EventName           string    `gorm:"column:event_name;type:varchar(63);not null" json:"eventName"`
-	ClusterID           string    `gorm:"column:cluster_id;type:uuid;not null" json:"clusterId"`
-	LeafHubName         string    `gorm:"size:256;not null" json:"-"`
+	ClusterName         string    `gorm:"column:cluster_name;type:varchar(63);not null" json:"clusterName"`
+	ClusterID           string    `gorm:"column:cluster_id;type:text" json:"clusterId"`
```
clyang82, line 56: we may add back `not null`

clyang82 (approved):
LGTM

PR #935, opened 2024-05-22
:seedling: Define the Kafka's data rentention policy explicity

yanmxa:
## Summary

`"log.retention.ms": 604800000`: the logs in each partition older than 7 days will be deleted
`"log.retention.bytes": 1073741824`: make sure each partition size is below 1 GB
`"log.segment.bytes": 107374182 `: each segment log in a partition will make sure the file size is 100 MB, this a not a deletion policy.

Either the "log.retention.ms" and "log.retention.bytes" condition is met. It will apply the deletion policy on the related partitions

## Related issue(s)

Fixes #


yanmxa:
The default data retention policy is:
```bash
log.retention.hours=168
```

data keeps 7 days

PR #948, opened 2024-06-03
:sparkles: add generate history.local_compliance data script

ldpliu:
## Related issue(s)

Fixes #


doc/simulation/local-policies/history_local_compliance.sql
```diff
@@ -0,0 +1,33 @@
+-- This script will generate <365> days data in history.local_compliance
+do
+$$
+declare
+    all_days int = 365;
```
yanmxa, line 5: Maybe we can convert the `all_days`(simulation_days) into a parameter. so we can run the SQL function more flexible！

PR #947, opened 2024-06-03
:bug: change default time of dashboards

ldpliu:
## Related issue(s)

Fixes #

PR #952, opened 2024-06-03
:bug: Fix dashboard error

clyang82:
## Summary

We add `cluster_name` in event.managed_cluster table, so both tables have `cluster_name`. that is why report `column reference "cluster_name" is ambiguous`
The fix is to specify the table name to have fixed cluster_name.

## Related issue(s)

image

Fixes #


clyang82:
dashboards change won't be test in ut and e2e. the errors are not related with this PR. it will be handled separately.
/override ci/prow/test-e2e

clyang82:
it was fixed in https://github.com/stolostron/multicluster-global-hub/pull/947

PR #955, opened 2024-06-04
:book: Add postgres perf

ldpliu:
## Related issue(s)
https://issues.redhat.com/browse/ACM-11597


ldpliu:
/hold
Need add simulate script when the that pr merged.

clyang82 (commented):
LGTM

doc/simulation/scenario/Scenario5: postgres_metrics.md
```diff
@@ -0,0 +1,129 @@
...
+- 1500 Managed Clusters
+- 250 Policies, 75,000 Replicated Policies
+
+## Environment
+
+1. Install the global hub and then join the 5 simulated managed hubs(Each hub has 300 managedclusters and 50 policies) into it
+
+2. Deploy the `multicluster-global-hub-agent` to the `hub1` ~ `hub5` cluster
+
+3. Generate 180/360/540 days data in `history.local_compliance`
```
yanmxa, line 15: A more realistic approach is generating 180/360/540 days in the local_status.compliance. And then use the [script](https://github.com/stolostron/multicluster-global-hub/blob/c1121dc263a1bba9f882519f3028774238bef090/manager/pkg/cronjob/task/local_compliance_history.go#L89-L115) to migrate compliance from status to history. 

It will keep the data not only on history but also in the status compliance table, the resource will be different from the current resource. also, we can see the data migration impacts on the postgres database. 

If it's time-consuming, we can add TODO in step 3. So I can merge this PR.
@ldpliu What do you think of it?

ldpliu: Currently, our script generate history data from local_status.compliance https://github.com/stolostron/multicluster-global-hub/blob/c1121dc263a1bba9f882519f3028774238bef090/doc/simulation/local-policies/history_local_compliance.sql#L28.

PR #956, opened 2024-06-04
:bug: Make the event between client-go and library-go can be compatible

yanmxa:
## Summary

**Note 1:**
The policy leverages the [library-go](https://github.com/openshift/library-go/blob/master/pkg/operator/events/recorder.go#L221-L237) to send events. In comparison, the placement/managed cluster uses the [client-go](https://github.com/kubernetes/client-go/blob/master/tools/events/event_recorder.go#L91-L113) to broadcast event. 
They both have different formats, like count, lastTimestamp, and source. This PR will make sure the agent can handle both events

**Note 2:**
The same behavior of the involved object happens a lot of times. It might just be with 1 event. and the changes will be the count increment and the lastTimestamp updated. As the issue described: https://issues.redhat.com/browse/ACM-11992. 
So we will use the lastTimestamp to replace the creatTimestamp which might keep the same value when the count is updated.

## Related issue(s)

Fixes #https://issues.redhat.com/browse/ACM-11992.

### DIFF
--- a/agent/pkg/status/controller/event/local_root_policy_emitter.go
+++ b/agent/pkg/status/controller/event/local_root_policy_emitter.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	cloudevents "github.com/cloudevents/sdk-go/v2"
 	"github.com/go-logr/logr"
@@ -24,6 +25,8 @@ import (
 )
 
 var _ generic.ObjectEmitter = &localRootPolicyEmitter{}
+
+var DeltaInterval = 10 * time.Second
 
 type localRootPolicyEmitter struct {
 	ctx             context.Context
@@ -76,8 +79,7 @@ func policyEventPredicate(ctx context.Context, name string, obj client.Object, c
 		return nil, false
 	}
 
-	// if it's a older event, then return false
-	if !filter.Newer(name, evt.CreationTimestamp.Time) {
+	if !filter.NewerWithNoise(name, getEventLastTime(evt).Time, DeltaInterval) {
 		return nil, false
 	}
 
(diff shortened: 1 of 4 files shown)


agent/pkg/status/controller/event/local_root_policy_emitter.go
```diff
@@ -191,3 +193,32 @@ func getInvolvePolicy(ctx context.Context, c client.Client, evt *corev1.Event) (
...
+
+// the client-go event: https://github.com/kubernetes/client-go/blob/master/tools/events/event_recorder.go#L91-L113
+// the library-go event: https://github.com/openshift/library-go/blob/master/pkg/operator/events/recorder.go#L221-L237
+func getEventLastTime(evt *corev1.Event) metav1.Time {
+	lastTime := evt.CreationTimestamp
+	if !evt.LastTimestamp.IsZero() {
+		lastTime = evt.LastTimestamp
+	}
+	if evt.Series != nil {
+		lastTime = metav1.Time(evt.Series.LastObservedTime)
```
clyang82, line 205: Do we need to check !evt.Series.LastObservedTime.isZero() here?

agent/pkg/status/controller/filter/time_filter.go
```diff
@@ -39,6 +39,16 @@ func Newer(key string, val time.Time) bool {
 	return val.After(old)
 }
 
+// NewerWithNoise compares the val time with cached the time, if not exist, then return true
+// Set a delta interval value(noise) to avoid filtering out some out-of order events that still need to be sent
```
clyang82, line 43: what do you mean `out-of-order` events here? can you provide an example? Thanks.

agent/pkg/status/controller/event/local_root_policy_emitter.go
yanmxa: From the [source code](https://github.com/kubernetes/client-go/blob/master/tools/events/event_broadcaster.go#L181-L188), we can see that if the series is not nil, The `LastObervedTime` will always be set. So I think we don't need to check it~

agent/pkg/status/controller/filter/time_filter.go
yanmxa: 1. kube watch events     managed hub(multicluster-global-hub):         filter                                  configmap
managed cluster2:  event2            1:56:20                                         -> 1:56:20                              1:56:20
managed cluster1:   event1            1:56:19                                                                                          1:56:19

2. kube watch events     managed hub(multicluster-global-hub): 
managed cluster2:  event2            1:56:20                       noise 10    -> 1:56:10                               1:56:20
managed cluster1:   event1            1:56:19                                                                                          1:56:19

clyang82 (approved):
LGTM

PR #958, opened 2024-06-05
:bug: clean strimi resources when delete mgh

ldpliu:
## Related issue(s)
https://issues.redhat.com/browse/ACM-10754

### DIFF
--- a/operator/pkg/controllers/hubofhubs/globalhub_prune.go
+++ b/operator/pkg/controllers/hubofhubs/globalhub_prune.go
@@ -5,19 +5,24 @@ import (
 	"fmt"
 	"time"
 
+	kafkav1beta2 "github.com/RedHatInsights/strimzi-client-go/apis/kafka.strimzi.io/v1beta2"
 	"github.com/go-logr/logr"
+	subv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	promv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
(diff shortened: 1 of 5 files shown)


operator/pkg/controllers/hubofhubs/globalhub_prune.go
```diff
@@ -189,6 +203,88 @@ func (r *MulticlusterGlobalHubReconciler) pruneNamespacedResources(ctx context.C
...
+	klog.Infof("Remove strimzi resources")
+	kafka := &kafkav1beta2.Kafka{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      transporter.KafkaClusterName,
+			Namespace: utils.GetDefaultNamespace(),
+		},
+	}
+	klog.Infof("Delete kafka cluster %v", kafka.Name)
+
+	if err := r.Client.Delete(ctx, kafka); err != nil && !errors.IsNotFound(err) {
```
clyang82, line 216: It appears that it is not operating as your expected. the deletion order should be `kafkatopic`/`kafkauser` -> `kafka` ->   `csv` -> `subscription`.

ldpliu: I tested it in my env. And the deletion order is no matter for kafkatopic/kafkauser/kafka.
Any way, I changed the order in current code for your suggestion.

clyang82: if the kafka CR was deleted, the entity operator can be deleted. the entity operator is responsible for cleanup the kafkatopic and kafkauser. You did not find issues during your test, that is because you delete those resources in the same time. but anyway, it is reasonable to comply with that order.

clyang82 (approved):
LGTM

PR #979, opened 2024-06-18
:sparkles: manager integration

yanmxa:
## Related issue(s)

Fixes #


manager/pkg/cronjob/scheduler_test.go
```diff
@@ -31,7 +31,7 @@ func TestScheduler(t *testing.T) {
 	var err error
 	testenv = &envtest.Environment{
```
clyang82, line 32: maybe move it to integration

yanmxa: Done.

PR #1005, opened 2024-07-01
kafka can not be deployed in latest env

ldpliu:
## Summary
I try to use make deploy to deploy globalhub env, but the kafka related resources can not be created.

## Related issue(s)

Fixes #


yanmxa:
We cannot remove the resource detection for the kafka resource. If the Kafka resources/crds are not created, then the strimiz cluster shouldn't be deployed.

yanmxa:
Maybe we can move it to here: https://github.com/stolostron/multicluster-global-hub/pull/1005/files#diff-7474661d81c6705ca2b4a20e4024ac153dd73a8e957cc14812198df5ffe00c98L188?

ldpliu:
> Maybe we can move it to here: https://github.com/stolostron/multicluster-global-hub/pull/1005/files#diff-7474661d81c6705ca2b4a20e4024ac153dd73a8e957cc14812198df5ffe00c98L188?

It should be OK now. Please help to  review again.

yanmxa:
we also need check this for release-2.11

ldpliu:
> we also need check this for release-2.11

2.11 do not have this issue, it's involved in main code

PR #1008, opened 2024-07-03
[ Release 2.11 ] upgrade oauth proxy for ocp4.16+

yanmxa:
## Related issue(s)
Reference: https://github.com/stolostron/multiclusterhub-operator/pull/778
Fixes #

### DIFF
--- a/operator/pkg/controllers/hubofhubs/globalhub_manager.go
+++ b/operator/pkg/controllers/hubofhubs/globalhub_manager.go
@@ -108,11 +108,16 @@ func (r *MulticlusterGlobalHubReconciler) reconcileManager(ctx context.Context,
 		return fmt.Errorf("failed to get the electionConfig %w", err)
 	}
 
+	oauthProxyImage, err := config.GetOauthProxyImage(ctx, r.Manager.GetClient())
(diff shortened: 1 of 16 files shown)


operator/pkg/controllers/hubofhubs/globalhub_manager.go
```diff
@@ -108,11 +108,16 @@ func (r *MulticlusterGlobalHubReconciler) reconcileManager(ctx context.Context,
 		return fmt.Errorf("failed to get the electionConfig %w", err)
 	}
 
+	oauthProxyImage, err := config.GetOauthProxyImage(ctx, r.Manager.GetClient())
```
clyang82, line 111: why do we need oauth-proxy in global hub manager?

yanmxa: We use the oauth-proxy for the authentication of the manager non-k8s API.

clyang82: I see. Thanks.

clyang82 (approved):
LGTM

PR #1009, opened 2024-07-03
upgrade strimzi to latest

ldpliu:
## Related issue(s)
https://issues.redhat.com/browse/ACM-12540
Fixes #


yanmxa:
Is the upgrade causing e2e unstable?

ldpliu:
> Is the upgrade causing e2e unstable?

I don't think so.
The first retest retest the intergration, and the next one retest the e2e, but it failed by kind cluster can not get csr.

clyang82 (commented):
Can we use other way to install kafka? for example:
```
${kubectl} -n strimzi create -f 'https://strimzi.io/install/latest?namespace=strimzi'
wait_command "${kubectl} -n strimzi get deploy strimzi-cluster-operator"
```
in this way, we do not need to keep CRD in our end. we just need to create kafkacluster and kafkatopics. @yanmxa WDYT?

ldpliu:
> Can we use other way to install kafka? for example:
> 
> ```
> ${kubectl} -n strimzi create -f 'https://strimzi.io/install/latest?namespace=strimzi'
> wait_command "${kubectl} -n strimzi get deploy strimzi-cluster-operator"
> ```
> 
> in this way, we do not need to keep CRD in our end. we just need to create kafkacluster and kafkatopics. @yanmxa WDYT?

Where do you get this link? `https://strimzi.io/install/latest?namespace=strimzi`, How to specify the 0.40.0 strimzi?

yanmxa:
> Can we use other way to install kafka? for example:
> 
> ```
> ${kubectl} -n strimzi create -f 'https://strimzi.io/install/latest?namespace=strimzi'
> wait_command "${kubectl} -n strimzi get deploy strimzi-cluster-operator"
> ```
> 
> in this way, we do not need to keep CRD in our end. we just need to create kafkacluster and kafkatopics. @yanmxa WDYT?

It seems this way only installs the latest strimzi operator. But we need to fix it, right?

test/script/util.sh
```diff
@@ -10,8 +10,8 @@ export ROUTE_VERSION=release-4.12
 export GO_VERSION=go1.21.7
 export GINKGO_VERSION=v2.17.2
 # Image
-export KAFKA_OPERATOR_IMG="quay.io/strimzi/operator:0.38.0"
-export KAFKA_IMG="quay.io/strimzi/kafka:0.38.0-kafka-3.6.0"
+export KAFKA_OPERATOR_IMG="quay.io/strimzi/operator:0.40.0"
```
clyang82, line 13: maybe use `latest` tag. the purpose here is to load the image at the beginning, because we found sometimes the image cannot be loaded successfully and need retry.

clyang82 (approved):
LGTM

PR #1022, opened 2024-07-16
disable the replicated policy event

yanmxa:
## Summary

The replicated policy events are duplicated and do not consist of the ACM dashboard.

## Related issue(s)

Fixes #


clyang82 (approved):
LGTM

Issue #1019, opened 2024-07-12
E2E test log 


yanmxa:
The log cannot appear for the following reason:

```bash
This file contained potentially sensitive information and has been removed.
```

clyang82:
I am working with security team to fix this issue.

PR #1024, opened 2024-07-18
:sparkles: Separate global hub from acm

clyang82:
## Summary

Separate global hub from acm so that the operator can be working without ACM installed.

## Related issue(s)

Fixes #


clyang82:
fixed: https://github.com/stolostron/multicluster-global-hub/issues/1019

operator/pkg/controllers/crd/crd_controller.go
```diff
@@ -69,9 +75,14 @@ func (r *CrdController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 	// the reconcile will update the resources map in multiple goroutines simultaneously
 	r.mu.Lock()
 	defer r.mu.Unlock()
+	// set resource as ready
 	r.resources[req.Name] = true
-	if !r.readyToWatchACMResources() {
-		return ctrl.Result{}, nil
```
yanmxa, line 74: We can check the ACM resource with `readyToWatchACMResources()` here. If false, then return. Don't need to check it each time on the following assertions.

operator/pkg/controllers/hubofhubs/controller.go
```diff
@@ -96,6 +101,385 @@ func NewGlobalHubController(mgr ctrl.Manager, addonMgr addonmanager.AddonManager
...
+	kubeClient kubernetes.Interface, operatorConfig *config.OperatorConfig,
+) (controller.Controller, error) {
+	globalHubController, err := controller.New(operatorconstants.GlobalHubControllerName, mgr, controller.Options{
+		Reconciler: NewGlobalHubReconciler(mgr, kubeClient, operatorConfig),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	if err := addGlobalHubControllerWatches(mgr, globalHubController); err != nil {
```
yanmxa, line 114: I saw you launch the global hub controller out of the crd checking. https://github.com/stolostron/multicluster-global-hub/pull/1024/files#diff-f7893cd19dbd21bb6025ab8336c0fa1799f30e7cfcd17fd0cbcb73c7c0d1e3b9R73

I'm afraid it may crash with `no kind find` error or something like that. Do you have tried to verify if it can work without these crd created? or is there any test cases to cover it?

operator/pkg/controllers/hubofhubs/controller.go
```diff
@@ -179,16 +576,18 @@ func (r *GlobalHubController) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, nil
 	}
 
-	// update the managed hub clusters
-	// only reconcile once: upgrade
-	if !r.upgraded {
-		if err = utils.RemoveManagedHubClusterFinalizer(ctx, r.Client); err != nil {
-			return ctrl.Result{}, fmt.Errorf("failed to upgrade from release-2.10: %v", err)
+	if config.IsACMResourceReady() {
```
yanmxa, line 579: If we must check the resource before launching the controller, described as this: https://github.com/stolostron/multicluster-global-hub/pull/1024/files#r1687316221

then we don't need to check acm resources here.

operator/pkg/controllers/hubofhubs/controller.go
```diff
@@ -201,244 +600,30 @@ func (r *GlobalHubController) Reconcile(ctx context.Context, req ctrl.Request) (
...
 		return ctrl.Result{}, err
 	}
 
-	if controllerutil.AddFinalizer(mgh, constants.GlobalHubCleanupFinalizer) {
-		if err = r.Client.Update(ctx, mgh, &client.UpdateOptions{}); err != nil {
-			if errors.IsConflict(err) {
-				r.log.Info("conflict when adding finalizer to mgh instance", "error", err)
-				return ctrl.Result{Requeue: true}, nil
-			}
+	if config.IsACMResourceReady() {
```
yanmxa, line 613: Same as: https://github.com/stolostron/multicluster-global-hub/pull/1024/files#r1687318905

manager/cmd/manager/main.go
```diff
@@ -250,38 +244,47 @@ func createManager(ctx context.Context,
...
 	if err != nil {
 		return nil, fmt.Errorf("failed to init spec transport bridge: %w", err)
 	}
-	if managerConfig.EnableGlobalResource {
-		if err := specsyncer.AddGlobalResourceSpecSyncers(mgr, managerConfig, producer); err != nil {
-			return nil, fmt.Errorf("failed to add global resource spec syncers: %w", err)
+
+	// TODO: refactor the manager to start the conflation manager so that it can handle the events from restful API
+
+	if managerConfig.WithACM {
```
yanmxa, line 255: If without the ACM, then we can return the manager directly. To avoid wrapping the following codes in a large block and reduce the number of nested if statements.

operator/pkg/controllers/hubofhubs/controller.go
clyang82: That is key for this PR. in `addGlobalHubControllerWatches` method, I have added the watcher for some resources which are existing in OCP environment. You can find I have added ACM specific resources in `crd_controller.go`. refer to this PR to dynamically add watcher - https://github.com/kubernetes-sigs/controller-runtime/pull/2881

yanmxa: Thanks! I see..., It goes deeper: to make the watch dynamic rather than the controller.

operator/pkg/controllers/hubofhubs/controller.go
yanmxa: All the watches have the same controller, so we need the if statements to check if the specific watches are ready for the related logic

operator/pkg/controllers/hubofhubs/controller.go
yanmxa: Done.

Issue #951, opened 2024-06-03
Store the offset in status.transport table for event


clyang82:
currently, if the event is failed to insert to database, the event will be ignore by the manager. it is better to store the offset in status.transport table so that it can be handled next time.

/assign @yanmxa

PR #1027, opened 2024-07-22
Add the event offset to database

yanmxa:
## Related issue(s)

Fixes https://github.com/stolostron/multicluster-global-hub/issues/951


manager/pkg/statussyncer/conflator/element_delta.go
```diff
@@ -61,6 +64,7 @@ func (e *deltaElement) Predicate(eventVersion *version.Version) bool {
 
 func (e *deltaElement) AddToReadyQueue(event *cloudevents.Event, metadata ConflationMetadata, cu *ConflationUnit) {
 	cu.readyQueue.DeltaEventJobChan <- NewConflationJob(event, metadata, e.handlerFunction, cu)
+	e.metadata = metadata
```
clyang82, line 67: will `metadata` be stored in DB?

yanmxa: This metadata is an interface, it's not the kube metadata. We only store the transport metadata(topic: offset) in the status. transport table

clyang82 (approved):
LGTM

PR #1032, opened 2024-07-26
:sparkles: Configure the topics by Global Hub operand

yanmxa:
## Summary

For the BYO case, the customer needs to create the configured topics or the default topic `gh-spec` and `gh-event` on their kafka cluster.

## Related issue(s)

Fixes # https://issues.redhat.com/browse/ACM-12954
References: https://issues.redhat.com/browse/ACM-11625


operator/apis/v1alpha4/multiclusterglobalhub_types.go
```diff
@@ -179,6 +179,21 @@ type PostgresConfig struct {
 
 // KafkaConfig defines the desired state of kafka
 type KafkaConfig struct {
+	// SpecTopic is the topic to distribute workload from global hub to managed hubs. The default value is "gh-spec"
+	// +kubebuilder:default="gh-spec"
+	SpecTopic string `json:"specTopic"`
```
clyang82, line 184: I would suggest to use `topics` as a parent structure. and then put spec/events inside. for spec, it should be an optional. if we do not create `spec` topic at the beginning, what problems we may have? I asked this question because I did not see any requirements to use global resources so far. It should make sense to not create it at the beginning. if the user enables global resource, then we create it.

clyang82:
```
topics:
  event: gh-event
```
for native case, we can automatically add managed hub cluster name. `gh-event.mh1`
for BYO case, just use `gh-event`.
WDYT?

yanmxa: Event if we don't enable the global resources, we still need the `spec` topic for resync resources.

clyang82: Do we use `spec` topic for local spec?

yanmxa: Yes

operator/apis/v1alpha4/multiclusterglobalhub_types.go
```diff
@@ -179,11 +179,29 @@ type PostgresConfig struct {
...
+type KafkaTopics struct {
+	// SpecTopic is the topic to distribute workload from global hub to managed hubs. The default value is "gh-spec"
+	// +kubebuilder:default="gh-spec"
+	SpecTopic string `json:"specTopic,omitempty"`
+
+	// StatusTopic defines the topic for the agent to report events and statuses to the manager.
+	// For the built-in Kafka, the default topic is "gh-event.*", where "*" is replaced by the managed hub cluster's name.
+	// For example, if the hub cluster is named "hub1", the default StatusTopic for it will be "gh-event.hub1".
+	// For BYO Kafka, the default topic is simply "gh-event" for all the managed hubs.
+	// +kubebuilder:default="gh-event.*"
```
clyang82, line 201: should it work with a customized name? for example: `abc-event.*`

clyang82: What is valid string for topic name? Can we validate the topic name?

operator/pkg/config/condition.go
```diff
@@ -45,6 +45,12 @@ const (
 	CONDITION_MESSAGE_GRAFANA_AVAILABLE = "Multicluster Global Hub Grafana has been deployed"
 )
 
+// NOTE: the status of GrafanaInitialized can be True or False
```
clyang82, line 48: need update the comments

operator/apis/v1alpha4/multiclusterglobalhub_types.go
yanmxa: It is just a string that is used to indicate how the topics will look like:
1. For the manager it means to subscribe to all the topics with `gh-event` prefix, And the confluent consumer will subscribe to the topic "^gh-event.*"
2. For the agent it means its event topic is `gh-event.<managed-hub-name>`

operator/pkg/config/condition.go
yanmxa: Done

operator/apis/v1alpha4/multiclusterglobalhub_types.go
yanmxa: > should it work with a customized name? for example: `abc-event.*`

Yes. It works now. 
The latency of updating manifests takes too long, but it is unrelated to the global hub. I log an issue to the addon framework to trace it: https://github.com/open-cluster-management-io/addon-framework/issues/281

operator/apis/v1alpha4/multiclusterglobalhub_types.go
```diff
@@ -179,11 +179,29 @@ type PostgresConfig struct {
...
 	// StorageSize specifies the size for storage
 	// +optional
 	StorageSize string `json:"storageSize,omitempty"`
 }
 
+// KafkaTopics is the transport topics for the manager and agent to communicate
+type KafkaTopics struct {
+	// SpecTopic is the topic to distribute workload from global hub to managed hubs. The default value is "gh-spec"
+	// +kubebuilder:default="gh-spec"
+	SpecTopic string `json:"specTopic,omitempty"`
```
ldpliu, line 195: We may need to add validation for the topic string.
For example: 
1. It should only contain specific character like A-Z/-
2. The length should not too long
3. ...

yanmxa: Yes, the spec topic is compliance with the Kafka naming convention less than 255: https://kafka.apache.org/documentation/#basic_ops_add_topic

operator/apis/v1alpha4/multiclusterglobalhub_types.go
yanmxa: > What is valid string for topic name? Can we validate the topic name?

Thanks @clyang82 and @ldpliu! I added the validation for the Kafka topic configuration: https://github.com/stolostron/multicluster-global-hub/pull/1032/commits/0f338f36d2d94856e2de2bebffd29d5f7e22a69c#diff-f6eee4fd3c63fa43ad70562751a2e0ab9bc0e0eefa6c173c37f5dbe8754b6679R120

PR #1038, opened 2024-07-31
Introudce Containerfile.operator for Productized build

clyang82:
## Related issue(s)

Fixes https://issues.redhat.com/browse/ACM-12399

### DIFF
--- a/agent/Containerfile.operator
+++ b/agent/Containerfile.operator
@@ -1,0 +1,41 @@
+# Copyright Contributors to the Open Cluster Management project
+
+# Stage 1: build the target binaries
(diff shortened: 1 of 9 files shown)


agent/Containerfile.operator
```diff
@@ -0,0 +1,62 @@
+# Copyright (c) 2024 Red Hat, Inc.
```
yanmxa, line 1: So we will keep both `Dockerfile` and `Containerfile.operator` for each component?

Is the `Containerfile.operator` only for the conflux pipeline?

clyang82: Dockerfile is a community build, while Containerfile.operator is for productized build.

PR #1041, opened 2024-08-02
E2E logs for the operator and manager

yanmxa:
## Related issue(s)

Fixes #

### DIFF
--- a/test/script/e2e_prow.sh
+++ b/test/script/e2e_prow.sh
@@ -37,6 +37,7 @@ trap 'echo "An error occurred" >&2; log_component_logs; exit 1;' ERR
 log_component_logs() {
     ssh "${OPT[@]}" "$HOST" "cd $HOST_DIR && . test/script/env.list && make e2e-log/operator" > >(tee "$ARTIFACT_DIR/e2e-operator.log")
     ssh "${OPT[@]}" "$HOST" "cd $HOST_DIR && . test/script/env.list && make e2e-log/manager" > >(tee "$ARTIFACT_DIR/e2e-manager.log")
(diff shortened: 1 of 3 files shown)


test/script/e2e_prow.sh
```diff
@@ -8,30 +8,54 @@ chmod 400 "${KEY}"
...
+
+echo "Setting up e2e environment..."
+setup_e2e
 
-ssh "${OPT[@]}" "$HOST" sudo yum install git wget jq -y
-# Insufficient resources creating kind clusters, modify parameters to expand
-ssh "${OPT[@]}" "$HOST" "sudo sh -c 'echo \"fs.inotify.max_user_watches=524288\" >> /etc/sysctl.conf && \
-                                     echo \"fs.inotify.max_user_instances=8192\" >> /etc/sysctl.conf && \
-                                     sysctl -p /etc/sysctl.conf'"
+trap 'echo "An error occurred" >&2; log_component_logs; exit 1;' ERR
```
clyang82, line 55: we do not install global hub in `setup_e2e`. I think the gather the logs should be right after `e2e-test-all` and before `e2e-test-prune`

yanmxa: If the e2e-test-all fails, it will not execute the logs if we put it between `e2e-test-all` and `e2e-test-prune`

The components are only installed when running the e2e. So the trap is set after deploying the e2e environment.

clyang82: but the global hub is installed in `e2e-test-all` case, how do you get the logs after deployed the e2e environment?

yanmxa: I use the `trap` tool to catch the error signal from `e2e-test-all`. If it catches, then get the logs of the components. 

So we can just  put the logic in front of the `e2e-test-all`

clyang82 (approved):
LGTM

PR #1043, opened 2024-08-02
:sparkles: Add the cache filter with topic prefix

yanmxa:
## Summary

The previous cache key is based on the `eventType`. But if we support renaming the topic, we need the agent to send the event to the new topic again after the topic is changed. Otherwise, it might lose data when renaming the topics

## Related issue(s)

Reference: https://github.com/stolostron/multicluster-global-hub/pull/1032

### DIFF
--- a/agent/pkg/status/controller/controller.go
+++ b/agent/pkg/status/controller/controller.go
@@ -80,7 +80,8 @@ func AddControllers(ctx context.Context, mgr ctrl.Manager, agentConfig *config.A
 	}
 
 	// lunch a time filter, it must be called after filter.RegisterTimeFilter(key)
-	if err := filter.LaunchTimeFilter(ctx, mgr.GetClient(), agentConfig.PodNameSpace); err != nil {
+	if err := filter.LaunchTimeFilter(ctx, mgr.GetClient(), agentConfig.PodNameSpace,
+		agentConfig.TransportConfig.KafkaConfig.Topics.StatusTopic); err != nil {
 		return fmt.Errorf("failed to launch time filter: %w", err)
 	}
 	return nil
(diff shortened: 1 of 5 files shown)


yanmxa:
The grafana isn't ready
```bash
❯ oc logs multicluster-global-hub-grafana-7f589bfdc9-2lcv2  -n multicluster-global-hub -c grafana-proxy
2024/08/05 14:06:00 provider.go:128: Defaulting client-id to system:serviceaccount:multicluster-global-hub:multicluster-global-hub-grafana
2024/08/05 14:06:00 provider.go:133: Defaulting client-secret to service account token /var/run/secrets/kubernetes.io/serviceaccount/token
2024/08/05 14:06:00 provider.go:351: Delegation of authentication and authorization to OpenShift is enabled for bearer tokens and client certificates.
2024/08/05 14:06:00 oauthproxy.go:203: mapping path "/" => upstream "http://localhost:3001/"
2024/08/05 14:06:00 oauthproxy.go:230: OAuthProxy configured for  Client ID: system:serviceaccount:multicluster-global-hub:multicluster-global-hub-grafana
2024/08/05 14:06:00 oauthproxy.go:240: Cookie settings: name:_oauth_proxy secure(https):true httponly:true expiry:12h0m0s domain:<default> samesite: refresh:after 8h0m0s
2024/08/05 14:06:00 oauthproxy.go:252: WARN: Configured to pass client specified bearer token upstream.
      Only use this option if you're sure that the upstream will not leak or abuse the given token.
      Bear in mind that the token could be a long lived token or hard to revoke.
2024/08/05 14:06:00 http.go:61: HTTP: listening on 127.0.0.1:4180
2024/08/05 14:06:00 http.go:107: HTTPS: listening on [::]:9443
I0805 14:06:00.522972       1 dynamic_serving_content.go:130] Starting serving::/etc/tls/private/tls.crt::/etc/tls/private/tls.key

```

clyang82 (approved):
LGTM

PR #1044, opened 2024-08-06
:bug: Update the manifestwork with addon manager

yanmxa:
## Summary

We should always use the same addon manager(which adds the addon agent) to update the manifests. Otherwise, it might not be triggerred to update

## Related issue(s)

Fixes # https://github.com/open-cluster-management-io/addon-framework/issues/281

### DIFF
--- a/operator/pkg/config/multiclusterglobalhub_config.go
+++ b/operator/pkg/config/multiclusterglobalhub_config.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"open-cluster-management.io/addon-framework/pkg/addonmanager"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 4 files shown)


operator/pkg/config/multiclusterglobalhub_config.go
```diff
@@ -80,8 +81,17 @@ var (
 	statisticLogInterval  = "1m"
 	metricsScrapeInterval = "1m"
 	imagePullSecretName   = ""
+	addonMgr              addonmanager.AddonManager
 )
 
+func SetAddonManager(addonManager addonmanager.AddonManager) {
```
clyang82, line 87: Can we create the addonmanager in `main.go` and then pass it in?

yanmxa: No. The [addon controller](https://github.com/stolostron/multicluster-global-hub/blob/main/operator/pkg/controllers/addon/addon_controller.go#L62)(which generates/contains the addon manager) is created by the crd controller. Which means we can get the addon manager only if the ACM resources is ready and the addon controller is created

clyang82 (approved):
LGTM

PR #1055, opened 2024-08-15
:bug: Disable the the topic deletion when removing the global hub agent

yanmxa:
### Kafka topic deletion:
1. Consumer Impact: The consumer typically detects the deletion during its polling process. The default polling interval (max.poll.interval.ms) is 5 minutes, meaning the consumer may not notice the deletion until the next poll.

2. Metadata Refresh: Kafka clients periodically refresh their metadata, which includes topic information. The frequency of this refresh (`metadata.max.age.ms`) can also affect how quickly the consumer becomes aware of the topic deletion. The default value is 5 minutes.

3. Rebalancing Trigger: When the Kafka consumer attempts to fetch data from a deleted topic, it will encounter an error like UnknownTopicOrPartitionException. This error can trigger a rebalance in the consumer group. The rebalance process redistributes the available partitions among the remaining topics to the consumers. The time it takes to complete the rebalance can vary based on the size of the consumer group and the number of partitions involved. Typically, rebalancing can take anywhere from a few seconds to a minute, depending on these factors.

### Kafka topic creation:

1. Metadata Discovery:

**Metadata Refresh Interval** (`metadata.max.age.ms`): The consumer discovers the new topic during its next metadata refresh, which can take up to 5 minutes by default. If this interval is reduced or a manual refresh is triggered, the discovery happens faster.

2. Rebalancing:

Once the new topic is discovered, the consumer group coordinator triggers a rebalance. During rebalancing, the coordinator redistributes the topic partitions among all the consumers in the group.
**Rebalance Time**: The rebalance process can take anywhere from a few seconds to a minute or more, depending on the size of the consumer group and the number of partitions.

## Related issue(s)

Fixes # https://issues.redhat.com/browse/ACM-13317

### DIFF
--- a/operator/pkg/controllers/addon/addon_installer.go
+++ b/operator/pkg/controllers/addon/addon_installer.go
@@ -134,11 +134,11 @@ func (r *AddonInstaller) reconclieAddonAndResources(ctx context.Context, cluster
 		}
 	}
 
-	// reconcile kafka resources
-	return ensureKafkaResource(cluster.Name)
+	// reconcile transport resources
+	return ensureTransportResource(cluster.Name)
 }
 
-func ensureKafkaResource(clusterName string) error {
+func ensureTransportResource(clusterName string) error {
 	// create kafka resource: user and topic
 	trans := config.GetTransporter()
 	if trans == nil {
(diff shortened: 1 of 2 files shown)


yanmxa:
@clyang82 @ldpliu

clyang82 (approved):
LGTM

PR #1061, opened 2024-08-19
upgrade kube version

ldpliu:
## Summary
The current controller-gen will throw panic when `make bundle`, and we should upgrade it.
## Related issue(s)

Fixes #


clyang82 (approved):
LGTM

PR #1064, opened 2024-08-20
fix kafkatopic can not be deleted

ldpliu:
## Summary
kafkatopic should be deleted before the kafkaoperator.

## Related issue(s)
https://issues.redhat.com/browse/ACM-13119
Fixes #

### DIFF
--- a/operator/pkg/controllers/addon/addon_controller_manifests.go
+++ b/operator/pkg/controllers/addon/addon_controller_manifests.go
@@ -94,19 +94,6 @@ type HohAgentAddon struct {
 	operatorConfig *config.OperatorConfig
 }
 
-func (a *HohAgentAddon) getMulticlusterGlobalHub() (*globalhubv1alpha4.MulticlusterGlobalHub, error) {
-	mghList := &globalhubv1alpha4.MulticlusterGlobalHubList{}
-	err := a.client.List(a.ctx, mghList)
-	if err != nil {
-		return nil, err
-	}
-	if len(mghList.Items) != 1 {
-		return nil, fmt.Errorf("the count of the mgh instance is not 1 in the cluster.%v", len(mghList.Items))
-	}
-
-	return &mghList.Items[0], nil
-}
-
 func (a *HohAgentAddon) installACMHub(cluster *clusterv1.ManagedCluster) bool {
 	if _, exist := cluster.GetLabels()[operatorconstants.GHAgentACMHubInstallLabelKey]; !exist {
 		return false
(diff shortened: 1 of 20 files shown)


operator/pkg/controllers/hubofhubs/prune/prune_reconciler.go
```diff
@@ -318,41 +318,52 @@ func (r *PruneReconciler) pruneManagedHubs(ctx context.Context) error {
...
-	klog.Infof("Delete kafkaUsers")
-
-	if err := r.Client.List(ctx, kafkaUserList, listOpts...); err != nil {
-		return err
-	}
-	for idx := range kafkaUserList.Items {
-		klog.Infof("Delete kafka user %v", kafkaUserList.Items[idx].Name)
-		if err := r.Client.Delete(ctx, &kafkaUserList.Items[idx]); err != nil && !errors.IsNotFound(err) {
-			return err
+	err := wait.PollUntilContextCancel(ctx, 5*time.Second, true, func(ctx context.Context) (bool, error) {
```
clyang82, line 324: I think it has problem to use `wait.PollUntilContextCancel` in reconcile method. it leads to have multiple process running. Can we just use reconciler's functionality to handle it? for example: if there is error, just return error so that it will be handled next time.

operator/pkg/controllers/addon/addon_controller_manifests.go
```diff
@@ -166,7 +153,7 @@ func (a *HohAgentAddon) GetValues(cluster *clusterv1.ManagedCluster,
 	if len(installNamespace) == 0 {
 		installNamespace = operatorconstants.GHAgentInstallNamespace
 	}
-	mgh, err := a.getMulticlusterGlobalHub()
+	mgh, err := utils.GetMulticlusterGlobalHub(a.ctx, a.client)
```
yanmxa, line 156: we already have a function in the config to retrieve the `mgh` cr. You can use is directly!

test/integration/utils/utils.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/stolostron/multicluster-global-hub/operator/api/operator/v1alpha4"
+)
+
+func DeleteMgh(ctx context.Context, runtimeClient client.Client, mgh *v1alpha4.MulticlusterGlobalHub) error {
```
yanmxa, line 28: we might need to cleanup the related resources, I'm afraid the force delete may not cleanup the others

operator/pkg/controllers/addon/addon_controller_manifests.go
ldpliu: the addon controller is a separate controller, and it should known the latest mgh status. we should not get it from config, as it may be not in real time.

test/integration/utils/utils.go
ldpliu: > we might need to cleanup the related resources, I'm afraid the force deletion may not cleanup the associated resources

It's only used to clean finalizer of mgh in integration test. there is not other resources should be cleaned here.

PR #1060, opened 2024-08-19
:sparkles: Add tranposrt config secret for agent and manager

yanmxa:
## Related issue(s)

Fixes #

PR #1054, opened 2024-08-13
:sparkles: Add transport secret for global hub manager and agent

yanmxa:
## Summary

The transport-config secret for both the manager and agent will look like this:

```yaml
apiVersion: v1
data:
  kafka.yaml: ...
kind: Secret
metadata:
  creationTimestamp: "2024-08-15T13:18:49Z"
  labels:
    global-hub.open-cluster-management.io/managed-by: global-hub-operator
    name: multicluster-global-hub-manager
  name: transport-config
  namespace: multicluster-global-hub
  ...
type: Opaque
```

### `kafka.yaml` for manager:

```yaml
bootstrap.server: 172.18.0.4:30095
topic.status: ^gh-event.*
topic.spec: gh-spec
cluster.id: ZP_ORdm-SBmK17lAFDUtQA
ca.key: ...
client.crt: ...
client.key: ...
```

### `kafka.yaml` for the Agent

- BYO | Standalone:
```yaml
bootstrap.server: 172.18.0.4:30095
topic.status: ^gh-event.*
topic.spec: gh-spec
cluster.id: ZP_ORdm-SBmK17lAFDUtQA
ca.key: ...
client.crt: ...
client.key: ...
```

- Built-in:

```yaml
bootstrap.server: 172.18.0.4:30095
topic.status: ^gh-event.*
topic.spec: gh-spec
cluster.id: ZP_ORdm-SBmK17lAFDUtQA
ca.secret: ...
client.secret: ...
```
## Related issue(s)

Fixes # https://issues.redhat.com/browse/ACM-13019

PR #1070, opened 2024-08-22
:bug: Provide the default name for BYO

clyang82:
## Summary

This is a temporary solution to fix this blocker. For BYO case, the ca and certs secret are still required. It can be removed in 
https://github.com/stolostron/multicluster-global-hub/pull/1060.
It was introduced in https://github.com/stolostron/multicluster-global-hub/pull/1054
That is why add a new test for BYO is so important for us @ldpliu 

## Related issue(s)

Fixes # https://issues.redhat.com/browse/ACM-13522


clyang82:
fixed: https://issues.redhat.com/browse/ACM-13198

PR #1090, opened 2024-09-06
:sparkles: Install the agent by helm chart

yanmxa:
## Related issue(s)

Fixes # https://issues.redhat.com/browse/ACM-12394


yanmxa:
The integration test failed caused the database condition isn't updated

/test test-integration

yanmxa:
/test test-e2e

application test error

ldpliu:
it looks good to me, @clyang82 Please help to take a look.

doc/standalone-agent/README.md
```diff
@@ -0,0 +1,75 @@
...
+```bash
+$ oc get secret transport-config -ojsonpath='{.data.kafka\.yaml}' | base64 -d
+bootstrap.server: kafka-...       
+topic.status: topic...           # topic for sending events
+ca.crt: LS0...
+client.crt: LS0...
+client.key: LS0...
+```
+
+### Step 3: Generate the `transport-config` Secret
```
clyang82, line 49: I do not think we need this step. the event-exporter is designed for sending events to customized Kafka. not for global hub kafka. if it is going to send data to global hub, do we need to install global hub agent in this cluster. that is something we need to think over. but for now, keep it simple.

doc/standalone-agent/Chart.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v2
+name: standalone-agent
```
clyang82, line 2: maybe we just use `event-exporter` as name. also for the folder name. what is your thoughts?

doc/standalone-agent/templates/deployment.yaml
```diff
@@ -1,29 +1,27 @@
...
-        - name: multicluster-global-hub-agent
-          image: quay.io/stolostron/multicluster-global-hub-agent:latest
+        - name: {{ .Release.Name }}
+          image: {{ .Values.image }}
           imagePullPolicy: Always
           args:
             - --zap-log-level=info
             - --pod-namespace=$(POD_NAMESPACE)
-            - --leaf-hub-name=event-exporter-cluster
+            - --leaf-hub-name={{ .Release.Name }}
```
clyang82, line 24: `--leaf-hub-name` may set as openshift ID. that is because if there are 2 agents installed with the same name, it is hard to distinguish from kafka message.

doc/standalone-agent/README.md
```diff
@@ -0,0 +1,75 @@
+
+
```
clyang82, line 2: Could you make a link to `dev-preview.md`? or just put it into `dev-preview.md`.

doc/standalone-agent/README.md
yanmxa: Okay, We can remove this step

doc/standalone-agent/Chart.yaml
yanmxa: It's ok to set the name as `event-exporter`

doc/standalone-agent/README.md
yanmxa: Done.

PR #1091, opened 2024-09-06
MGMT-18597: Poll Stackrox Central for violation counts and push it to Kafka

jhernand:
## Summary

This patch changes the agent so that it discovers Stackrox _central_ instances in the hub, polls them to extract the summary of security violations and sends them to the manager via the Kafka broker.

## Related issue(s)

https://issues.redhat.com/browse/MGMT-18597


operator/pkg/controllers/agent/manifests/templates/agent/multicluster-global-hub-agent-deployment.yaml
```diff
@@ -46,6 +46,8 @@ spec:
             - --qps={{.AgentQPS}}
             - --burst={{.AgentBurst}}
             - --enable-pprof={{.EnablePprof}}
+            - --acs-poll-interval=1m
+            - --acs-integration=true
```
clyang82, line 50: the value should be from `global-hub.open-cluster-management.io/with-stackrox-integration`

operator/pkg/controllers/agent/manifests/templates/agent/multicluster-global-hub-agent-deployment.yaml
```diff
@@ -46,6 +46,8 @@ spec:
             - --qps={{.AgentQPS}}
             - --burst={{.AgentBurst}}
             - --enable-pprof={{.EnablePprof}}
+            - --acs-poll-interval=1m
```
clyang82, line 49: Do you want to configure it in annotation for testing? or just use `1m` for dev preview.

agent/pkg/status/controller/security/stackrox_central_data_syncer.go
```diff
@@ -0,0 +1,303 @@
...
+		}
+	}()
+	<-ctx.Done()
+
+	a.log.Info("stopped stackrox data syncer")
+
+	return nil
+}
+
+func AddStackroxDataSyncer(
```
clyang82, line 288: How about calling `AddStackroxDataSyncer` in `reconcile` of  stackroxCentralCRDController ?

agent/pkg/status/controller/security/stackrox_central_cr_controller.go
```diff
@@ -0,0 +1,65 @@
...
+			reqLogger.Info("Deleting Stackrox Central instance")
+			delete(stackroxCentrals, stackroxCentral{name: request.Name, namespace: request.Namespace})
+		}
+		return ctrl.Result{}, nil
+	}
+
+	if _, ok := stackroxCentrals[stackroxCentral{name: request.Name, namespace: request.Namespace}]; !ok {
+		reqLogger.Info("Found a new central instance to sync")
+		stackroxCentrals[stackroxCentral{name: request.Name, namespace: request.Namespace}] = &stackroxCentralData{}
+	}
```
clyang82, line 40: Can we put the real data into `stackroxCentralData`? if so, we do not need to getresource in `stackroxDataSyncer`. in `stackroxDataSyncer`, just get the information from that map.

clyang82 (commented):
Thanks for your contribution. Append some comments. Please also add some tests to pass the sonarcloud check. Thanks.

operator/pkg/controllers/agent/manifests/templates/agent/multicluster-global-hub-agent-deployment.yaml
jhernand: Yes, I will address that.

operator/pkg/controllers/agent/manifests/templates/agent/multicluster-global-hub-agent-deployment.yaml
jhernand: I think it makes sense to have it as an annotation. For example `global-hub.open-cluster-management.io/with-stackrox-poll-interval: "1m"`. Makes sense?

agent/pkg/status/controller/security/stackrox_central_crd_controller.go
```diff
@@ -0,0 +1,85 @@
...
+func LaunchStackroxCentralCRDController(ctx context.Context, mgr ctrl.Manager) error {
+	controller := stackroxCentralCRDController{
+		ctx:    ctx,
+		mgr:    mgr,
+		log:    ctrl.Log.WithName("stackrox-central-crd-controller"),
+		client: mgr.GetClient(),
+	}
+
+	// initial reconcile
+	if _, err := controller.Reconcile(ctx, reconcile.Request{}); err != nil {
```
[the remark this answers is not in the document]
jhernand: I agree. Will check and remove it.

agent/pkg/status/controller/security/stackrox_central_data_syncer.go
jhernand: Yes, you are right. I think we have an unnecessary intermediate controller here. I am checking.

agent/pkg/status/controller/security/stackrox_central_cr_controller.go
jhernand: Yes, will do it.

agent/pkg/status/controller/security/stackrox_central_crd_controller.go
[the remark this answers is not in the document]
jhernand: @clyang82 I addressed this issue in a separate patch: https://github.com/stolostron/multicluster-global-hub/pull/1091/commits/e9cb38b8189e70a6d4a67fe232a91423ab9ad8e2, in case it simplifies your review. I will continue to add patches to address your other comments. Eventually, when the pull request is ready to merge, I can squash all of that.

operator/pkg/controllers/agent/manifests/templates/agent/multicluster-global-hub-agent-deployment.yaml
jhernand: This is addressed in this commit: https://github.com/stolostron/multicluster-global-hub/pull/1091/commits/64ab32684302bdc26cb3efe927b864c8ad397b02 .

operator/pkg/controllers/agent/manifests/templates/agent/multicluster-global-hub-agent-deployment.yaml
jhernand: This is addressed in this commit: https://github.com/stolostron/multicluster-global-hub/pull/1091/commits/177b44091ba70b32342fbb5e1500beeebe3e9fb3 .

agent/pkg/status/controller/security/stackrox_central_data_syncer.go
jhernand: This is addressed in this commit: https://github.com/stolostron/multicluster-global-hub/pull/1091/commits/50caf83702f222ad1b299ad22e0665304b73ed8a .

agent/pkg/status/controller/security/stackrox_central_cr_controller.go
jhernand: @clyang82  After checking I see that the reason that we do it this way is that we will refresh the data (in particular the token) not only the first time that we discover the StackRox `central` object, but also when we try to connect to the API and it returns a 401 or 403 (unauthorized or forbidden). That way when the token expires and the user updates the secret containing it we will automatically fetch it without having to wait for a resync of the controller.

clyang82: will the token be updated in `central` object? if so, it will always trigger a reconciler

jhernand: No, the `central` object will only contain an annotation referencing the secret that contains the token, something like this:

```yaml
apiVersion: platform.stackrox.io/v1alpha1
kind: Central
metadata:
  annotations:
    global-hub.open-cluster-management.io/with-stackrox-credentials-secret: multicluster-global-hub-agent/acs-token
```

So when the token is expired the user will need to change the secret, but not the `central` object. Instead of watching the secrets we think it is better to just react when there is a connection failure: at that point we go and fetch the connection details from the secret.

clyang82: Good to know that. Thanks for your explanation.

clyang82:
Do you mind add some UT to pass the code coverage check?

agent/cmd/agent/main.go
```diff
@@ -242,17 +246,15 @@ func transportCallback(mgr ctrl.Manager, agentConfig *config.AgentConfig,
...
-		},
-		&policyv1.Policy{}:                  {},
-		&clusterv1.ManagedCluster{}:         {},
-		&clustersv1alpha1.ClusterClaim{}:    {},
-		&routev1.Route{}:                    {},
-		&placementrulev1.PlacementRule{}:    {},
-		&clusterv1beta1.Placement{}:         {},
-		&clusterv1beta1.PlacementDecision{}: {},
-		&appsv1alpha1.SubscriptionReport{}:  {},
+		&apiextensionsv1.CustomResourceDefinition{}: {},
```
yanmxa, line 249: Why was the cache filter `Field: fields.OneTermEqualSelector("metadata.name", "clustermanagers.operator.open-cluster-management.io")` removed for the CRD resource?

This change will significantly increase the agent's footprint.

jhernand: The reason to remove this is that we want to detect the presence of the StackRox `centrals.platform.stackrox.io` CRD. Only when that is present we add the controller that watches the instances. Unfortunately there is no way (that I know) to configure the cache to watch for *two* names: field selectors don't support the `in` operator. So what we did is remove this restriction here and added explicit checks in the reconciler:

```go
func (c *crdController) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
	switch {
	case request.Name == clusterManagersCRDName:
		return c.reconcileClusterManagers(ctx, request)
	case request.Name == stackRoxCentralCRDName && c.agentConfig.EnableStackroxIntegration:
		return c.reconcileStackRoxCentrals()
	}
	return ctrl.Result{}, nil
}
```

So the previous behavior shouldn't be affected.

It is true that this means that the cache will store all the CRDs of the cluster instead of just one, which requires more memory. For example, in a cluster with ACM and StackRox it will require approx 7 MiB more. The alternative would be to create two separate caches, and therefore two separate controller managers: one for the current behaviour and another one just to check the presence of the StackRox CRD. Would you prefer that @yanmxa? Do you know any other alternative?

jhernand: @clyang82 @yanmxa did you consider using the `builder.OnlyMetadata` option to reduce the amount of memory used by the cache? You are already doing that in the CRD controller of the operator:

https://github.com/stolostron/multicluster-global-hub/blob/7476c3d96aa7810bee8d1b3b1ed4193b683b4292/operator/pkg/controllers/crd/crd_controller.go#L205-L209

Would that be acceptable also in the CRD controller of the agent? Instead of 7 MiB that would require 100 KiB.

jhernand:
> Do you mind add some UT to pass the code coverage check?

Yes, I am working on that.

agent/cmd/agent/main.go
yanmxa: Great! You found it! I would like to use the [watchMetadata](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/builder#TypedBuilder.WatchesMetadata) considering the extensibility!

jhernand: Maybe we can do that change in a separate pull request? https://github.com/stolostron/multicluster-global-hub/pull/1097

jhernand: That ^ pull request has been merged, so I assume this is no longer a concern.

clyang82 (approved):
LGTM

Issue #1031, opened 2024-07-23
E2E error reports


yanmxa:
It occasionally encounters the following error when installing the e2e environment.
```bash
...
kubectl cluster-info --context kind-hub2
Thanks for using kind! 😊
kind create cluster --name hub2 --wait 5m: Success.  
kubectl config rename-context kind-hub2 hub2  
 Attempt 1...  
Context "kind-hub2" renamed to "hub2".
kubectl config rename-context kind-hub2 hub2: Success.  
kubectl config set-cluster kind-hub2 --server=https://172.18.0.3:6443/  
 Attempt 1...  
error: open /tmp/multicluster-global-hub/test/script/config/clusters.lock: file exists
Set kubectl context to "kind-hub1"
You can now use your cluster with:
kubectl cluster-info --context kind-hub1
Not sure what to do next? 😅  Check out https://kind.sigs.k8s.io/docs/user/quick-start/
...
```

yanmxa:
https://github.com/stolostron/multicluster-global-hub/pull/1024

yanmxa:
https://github.com/stolostron/multicluster-global-hub/pull/1041

Add the the logs for further troubleshooting~

PR #1098, opened 2024-09-12
:bug: e2e kind cluster creating error

yanmxa:
## Related issue(s)

Fixes # https://github.com/stolostron/multicluster-global-hub/issues/1031

### DIFF
--- a/test/script/util.sh
+++ b/test/script/util.sh
@@ -106,7 +106,7 @@ kind_cluster() {
   dir="${CONFIG_DIR:-$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)}"
   local cluster_name="$1"
   local kubeconfig="$dir/$cluster_name"
-  while [ ! -f "$kubeconfig" ]; do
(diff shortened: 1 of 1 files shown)


test/script/util.sh
```diff
@@ -106,7 +106,7 @@ kind_cluster() {
   dir="${CONFIG_DIR:-$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)}"
   local cluster_name="$1"
   local kubeconfig="$dir/$cluster_name"
-  while [ ! -f "$kubeconfig" ]; do
+  while [ ! -f "$kubeconfig" ] || ! kubectl config get-contexts -o name | grep -wq "$cluster_name"; do
```
clyang82, line 109: Do you mean that we have the case to call `kind_cluster` when the cluster is created already?

yanmxa: No. When we call the `kind_cluster` to create the cluster, we also want to generate the `kubeconfig` and rename the `context` after the cluster is created. However, even if the cluster is created, They may not generate the `kubeconfig` of renaming the `context`. We must re-create the cluster to ensure they are ready.

clyang82: Did we have logic to re-create the cluster already?

PR #1101, opened 2024-09-12
:bug: Fix the issue of event loss caused by time filtering.

yanmxa:
## Related issue(s)

Fixes # https://issues.redhat.com/browse/ACM-14262

## Tests
* [x] Unit/function tests have been added and incorporated into `make unit-tests`.
* [x] Integration tests have been added and incorporated into `make integration-test`.
* [ ] E2E tests have been added and incorporated into `make e2e-test-all`.
* [ ] List other manual tests you have done.

PR #1105, opened 2024-09-15
:sparkles: Add 'source' column to security violations dashboard

danmanor:
## Summary

This PR adds a `source` column to the security violations dashboard for referencing the `central` instance described by the record.

This is how the dashboard will look like:

image

## Related issue(s)

https://issues.redhat.com/browse/MGMT-18903

Fixes #

supporting multiple ACS instances per hub

## Tests

Tested in a local environment


danmanor:
FYI - @oourfali

operator/pkg/controllers/hubofhubs/storage/database/2.tables.sql
```diff
@@ -150,6 +150,7 @@ CREATE TABLE IF NOT EXISTS security.alert_counts (
     high integer NOT NULL,
     critical integer NOT NULL,
     detail_url text NOT NULL,
+    source text NOT NULL,
```
jhernand, line 153: I think we will want to make the new `source` column part of the primary key, otherwise once a row for a hub has been inserted you will not be able to insert any other row for that same hub. So the table will need to look like this:

```sql
CREATE TABLE IF NOT EXISTS security.alert_counts (
    hub_name text NOT NULL,
    source text NOT NULL,
    ...,
    PRIMARY KEY (hub_name, source) 
);
```

As this version hasn't been released I think there is no need for a migration. But if a migration was needed then I think it should look like this:

```sql
TRUNCATE TABLE security.alert_counts;
ALTER TABLE security.alert_counts ADD COLUMN source NOT NULL;
ALTER TABLE security.alert_counts DROP CONSTRAINT alert_counts_pkey;
ALTER TABLE security.alert_counts ADD PRIMARY_KEY (hub_name, source);
```

Note that the table needs to be truncated because we don't know the value of the new _source_ column for existing rows. I think this isn't a problem because the values will be re-populated the next time that the ACS servers are polled.

@clyang82, @bjoydeep do we need to include a migration for this when the change hasn't been released yet?

test/integration/manager/status/security_alert_counts_handler_test.go
```diff
@@ -66,6 +69,7 @@ var _ = Describe("SecurityAlertCountsHandler", Ordered, func() {
 			g.Expect(high).To(Equal(3))
 			g.Expect(critical).To(Equal(4))
 			g.Expect(detailURL).To(Equal("https://hub1/violations"))
+			g.Expect(source).To(Equal("rhacs-operator/stackrox-central-services"))
 		}
 		Eventually(check, 30*time.Second, 100*time.Millisecond).Should(Succeed())
 	})
```
jhernand, line 75: Would be nice to add a test to:

1. Verify when we receive data from two different ACS instances in the same hub the result is that two rows are created in the table.
2. Verify that when we have already two rows in the table and we receive new data only the correct row is updated.

operator/pkg/controllers/hubofhubs/storage/database/2.tables.sql
danmanor: Thank you, I added the primary key change. Regarding the migration, I initially thought about adding it to, but then I figures the version isn't released yet so no reason to do it. lets see what @clyang82, @bjoydeep say

bjoydeep: Correct. Migration is not yet needed for the reasons you guys said above

pkg/database/models/security.go
```diff
@@ -25,6 +25,10 @@ type SecurityAlertCounts struct {
 	//	https://central-rhacs-operator.apps.../main/violations
 	DetailURL string `gorm:"column:detail_url;not null"`
 
+	// Source is the Central CR instance from which the data was retrieved.
+	// This should follow the format: "<namespace>/<name>"
+	Source string `gorm:"column:source;not null"`
```
bjoydeep, line 30: The primary key changes, right?

operator/pkg/controllers/hubofhubs/storage/database/2.tables.sql
bjoydeep: @jhernand @danmanor thinking about this -- isnt the `detail` column data going to be unqiue? if so, what happens if that is the unique instead of the hub_name? or what if `hub_name.detail` a concat primary key instead of adding a source column?

operator/pkg/controllers/hubofhubs/storage/database/2.tables.sql
```diff
@@ -144,12 +144,14 @@ CREATE TABLE IF NOT EXISTS status.transport (
 );
 
 CREATE TABLE IF NOT EXISTS security.alert_counts (
-    hub_name text PRIMARY KEY,
+    hub_name text NOT NULL,
     low integer NOT NULL,
     medium integer NOT NULL,
     high integer NOT NULL,
     critical integer NOT NULL,
     detail_url text NOT NULL,
```
bjoydeep, line 152: @jhernand @danmanor thinking about this -- isnt the detail column data going to be unqiue when we have multiple ACS central on 1 ACM hub? if so, what happens if that is the unique instead of the hub_name? or what if hub_name.detail a concat primary key instead of adding a source column?

jhernand: You are right that the detail URL should be unique @bjoydeep. But I think it would be good to have the `source` anyhow. We intend to put in the `source` column the namespace/name of the ACS central object, not only because it is unique but also because it will be useful for the user of the dashboard. For example, imagine that the user has different central instances with names like "production", "integration", "development". She will be able to go to the table and filter to see only the "production" violations, for example. That may be also possible with the detail URL, depending on the DNS naming that they use, but I think having the source is cleaner.

pkg/database/models/security.go
jhernand: Yes, the primary key is now composite: `(hub_name, source)`

operator/pkg/controllers/hubofhubs/storage/database/2.tables.sql
danmanor: I agree the URL should be unique, but I am not sure it is worth the change as both options are valid. WDYT @jhernand ?

jhernand: My preference is to add the `source` column.

clyang82 (approved):

clyang82:
Summarizing 1 Failure:
  [FAIL] SecurityAlertCountsHandler [It] Should be able to sync security alert counts event from multiple central instances in hub
  /go/src/github.com/stolostron/multicluster-global-hub/test/integration/manager/status/security_alert_counts_handler_test.go:175

test/integration/manager/status/security_alert_counts_handler_test.go
```diff
@@ -57,15 +73,194 @@ var _ = Describe("SecurityAlertCountsHandler", Ordered, func() {
...
+
+		time.Sleep(100 * time.Millisecond)
+
+		err = producer.SendEvent(ctx, *event2)
+		Expect(err).To(Succeed())
+
+		By("Check the table")
+		db := database.GetSqlDb()
+		Expect(db).ToNot(BeNil())
+		sql := fmt.Sprintf(
```
clyang82, line 129: @yanmxa Do you think the error shares the same root cause as https://github.com/stolostron/multicluster-global-hub/pull/1101?

clyang82: trace this issue in https://github.com/stolostron/multicluster-global-hub/issues/1109

yanmxa: refer to https://github.com/stolostron/multicluster-global-hub/issues/1109#issuecomment-2360653973

PR #1108, opened 2024-09-18
reduce the error in log

ldpliu:
## Related issue(s)

Fixes #

## Tests


operator/pkg/config/multiclusterglobalhub_config.go
```diff
@@ -357,6 +357,9 @@ func GetMulticlusterGlobalHub(ctx context.Context, c client.Client) (*v1alpha4.M
 	if err != nil {
 		return nil, err
 	}
+	if len(mghList.Items) == 0 {
+		return nil, nil
+	}
```
clyang82, line 362: why do we need this change? Cannot be covered in `len(mghList.Items) != 1`

operator/pkg/controllers/hubofhubs/controller.go
```diff
@@ -391,7 +391,7 @@ func watchMulticlusterGlobalHubPredict() predicate.TypedPredicate[*v1alpha4.Mult
 			return true
 		},
 		UpdateFunc: func(e event.TypedUpdateEvent[*v1alpha4.MulticlusterGlobalHub]) bool {
-			return e.ObjectOld.GetResourceVersion() != e.ObjectNew.GetResourceVersion()
+			return e.ObjectOld.GetGeneration() != e.ObjectNew.GetGeneration()
```
clyang82, line 394: we cannot change it to generation. in other controllers, we update the mcgh CR status in order to trigger a reconciler.

operator/pkg/controllers/hubofhubs/controller.go
```diff
@@ -604,16 +604,12 @@ func (r *GlobalHubReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	r.log.V(2).Info("reconciling mgh instance", "namespace", req.Namespace, "name", req.Name)
 	mgh, err := config.GetMulticlusterGlobalHub(ctx, r.client)
 	if err != nil {
-		if errors.IsNotFound(err) {
```
clyang82, line 607: This is a standard to check if the CR does not exist, just exit.

operator/pkg/controllers/hubofhubs/prune/prune_reconciler.go
```diff
@@ -406,14 +406,21 @@ func (r *PruneReconciler) pruneStrimziResources(ctx context.Context) error {
 			return err
 		}
 	}
+	// Wait kafkatopic is removed
+	err := wait.PollUntilContextTimeout(ctx, 1*time.Second, 5*time.Second, true, func(ctx context.Context) (bool, error) {
```
clyang82, line 410: Do not suggest to use `wait.PollUntilContextTimeout` in reconcile loop.

operator/pkg/controllers/hubofhubs/controller.go
ldpliu: > we cannot change it to generation. in other controllers, we update the mcgh CR status in order to trigger a reconciler.

It only impact the hubofhub controller, I did not found which controller need status to trigger it?
I think this controller should also do not be triggered by mgh stautus.

operator/pkg/controllers/hubofhubs/controller.go
ldpliu: > This is a standard to check if the CR does not exist, just exit.

This error returned from https://github.com/stolostron/multicluster-global-hub/pull/1108/files#diff-341f24952eb12c78df0cce96a0c8cf1cc29401fb3812de4968dc5f9ede5bdca9L605
It can not be isnotfound error, if mgh do not exist, we will return mgh is nil.

operator/pkg/config/multiclusterglobalhub_config.go
ldpliu: done

operator/pkg/controllers/hubofhubs/prune/prune_reconciler.go
ldpliu: done

operator/pkg/controllers/hubofhubs/controller.go
clyang82: Here is sample in operator-sdk - https://github.com/operator-framework/operator-sdk/blob/819984d4c1a51c8ff2ef6c23944554148ace0752/testdata/go/v4/memcached-operator/internal/controller/memcached_controller.go#L89

operator/pkg/controllers/hubofhubs/controller.go
clyang82: https://github.com/stolostron/multicluster-global-hub/blob/6b5f186b2cfee7d93f429b959d1fdb174993541b/operator/pkg/controllers/hubofhubs/transporter/protocol/strimzi_kafka_controller.go#L69

clyang82 (approved):
LGTM

yanmxa:
@ldpliu @ldpliu We cannot just return nil when the database of the kafka cluster is not ready. 
If it returns it nil, after a while, the database is ready. it will not trigger the following step, like installing the manager, or other components!

PR #1110, opened 2024-09-19
Have a migration controller to reconcile the managedclustermigation CR

clyang82:
## Related issue(s)

Fixes #

## Tests


morvencao (commented):
how do we delete migration CR? will it be left in the cluster after the managed cluster migration?

clyang82:
> how do we delete migration CR? will it be left in the cluster after the managed cluster migration?

They are 2 options:
1. remove the migration CR once the migration is done
2. keep it in the cluster so that we can check the status.
Need to think over. for dev preview, just leave it in the cluster.

PR #1121, opened 2024-09-23
:bug: Create inventory always

clyang82:
## Summary

Currently, only create `inventory` dbname when the CR has `with-inventory` annotation. the db won't be created if add `with-inventory` annotation latter because the postgres sts is already running, the init configmap won't be executed.
for the longer term, we will enable inventory by default. So, just create it regardless of whether we have this annotation or not.

## Related issue(s)

Fixes #

## Tests
Tested in local environment


yanmxa:
LGTM

PR #1123, opened 2024-09-23
:bug: Increment version in stackrox syncer

danmanor:
## Summary

Use one version object throughout `stackrox` syncing and increment it accordingly.

## Tests

### DIFF
--- a/agent/pkg/status/controller/security/stackrox_syncer.go
+++ b/agent/pkg/status/controller/security/stackrox_syncer.go
@@ -16,7 +16,6 @@ import (
 	"github.com/go-logr/logr"
 	routev1 "github.com/openshift/api/route/v1"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
 	cr "sigs.k8s.io/controller-runtime"
@@ -428,10 +427,10 @@ func (s *StackRoxSyncer) sync(ctx context.Context, data *stackRoxData) error {
 			return fmt.Errorf("failed to generate struct for kafka message: %v", err)
 		}
 
-		// If the payload not updated since the previous one, then don't need sync it again
-		if equality.Semantic.DeepEqual(messageStruct, s.lastSentData) {
-			return nil
-		}
+		// TODO: If the payload not updated since the previous one, then don't need sync it again
+		// if equality.Semantic.DeepEqual(messageStruct, s.lastSentData) {
+		// 	return nil
+		// }
 
 		s.currentVersion.Incr()
 		if err := s.produce(ctx, messageStruct); err != nil {
--- a/agent/pkg/status/controller/security/stackrox_syncer.go
+++ b/agent/pkg/status/controller/security/stackrox_syncer.go
@@ -11,6 +11,7 @@ import (
 	"sync"
 	"time"
 
+	cloudevents "github.com/cloudevents/sdk-go/v2"
 	cecontext "github.com/cloudevents/sdk-go/v2/context"
 	"github.com/go-logr/logr"
 	routev1 "github.com/openshift/api/route/v1"
@@ -22,8 +23,9 @@ import (
 	crmanager "sigs.k8s.io/controller-runtime/pkg/manager"
 
 	"github.com/stolostron/multicluster-global-hub/agent/pkg/clients"
-	"github.com/stolostron/multicluster-global-hub/agent/pkg/status/controller/generic"
+	"github.com/stolostron/multicluster-global-hub/agent/pkg/status/controller/config"
 	"github.com/stolostron/multicluster-global-hub/pkg/bundle/version"
+	eventversion "github.com/stolostron/multicluster-global-hub/pkg/bundle/version"
 	"github.com/stolostron/multicluster-global-hub/pkg/enum"
 	"github.com/stolostron/multicluster-global-hub/pkg/transport"
 )
@@ -48,15 +50,15 @@ type StackRoxSyncerBuilder struct {
 // StackRoxSyncer knows how to pull multiple StackRox API servers to extract information and send it to Kafka. Don't
 // create instances of this type directly, use the NewStackRoxSyncer function instead.
 type StackRoxSyncer struct {
-	logger       logr.Logger
-	topic        string
-	producer     transport.Producer
-	kubeClient   crclient.Client
-	pollInterval time.Duration
-	dataLock     *sync.Mutex
-	dataMap      map[types.NamespacedName]*stackRoxData
-	requests     []stackRoxRequest
-	version      *version.Version
+	logger         logr.Logger
+	topic          string
+	producer       transport.Producer
+	kubeClient     crclient.Client
+	pollInterval   time.Duration
+	dataLock       *sync.Mutex
+	dataMap        map[types.NamespacedName]*stackRoxData
+	requests       []stackRoxRequest
+	currentVersion *version.Version
 }
 
 // stackRoxConnDetails contains the details of the StackRox central.
@@ -140,20 +142,17 @@ func (b *StackRoxSyncerBuilder) Build() (result *StackRoxSyncer, err error) {
 		return
 	}
 
-	version := version.NewVersion()
-	version.Incr()
-
 	// Create and populate the object:
 	result = &StackRoxSyncer{
-		logger:       b.logger,
-		topic:        b.topic,
-		producer:     b.producer,
-		kubeClient:   b.kubeClient,
-		pollInterval: b.pollInterval,
-		dataLock:     &sync.Mutex{},
-		dataMap:      map[types.NamespacedName]*stackRoxData{},
-		requests:     stackRoxRequests,
-		version:      version,
+		logger:         b.logger,
+		topic:          b.topic,
+		producer:       b.producer,
+		kubeClient:     b.kubeClient,
+		pollInterval:   b.pollInterval,
+		dataLock:       &sync.Mutex{},
+		dataMap:        map[types.NamespacedName]*stackRoxData{},
+		requests:       stackRoxRequests,
+		currentVersion: version.NewVersion(),
 	}
 	return
 }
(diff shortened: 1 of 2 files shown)


danmanor:
FYI - @jhernand

test/integration/manager/status/security_alert_counts_handler_test.go
```diff
@@ -100,10 +100,9 @@ var _ = Describe("SecurityAlertCountsHandler", Ordered, func() {
 			DetailURL: DetailURL,
 			Source:    source1,
 		}
-		event1 := ToCloudEvent(leafHubName, string(enum.SecurityAlertCountsType), version1, dataEvent1)
+		event1 := ToCloudEvent(leafHubName, string(enum.SecurityAlertCountsType), version, dataEvent1)
```
yanmxa, line 104: Can you move the code to line 104
```go
By("Sync events with transport")
err := producer.SendEvent(ctx, *event1)
Expect(err).To(Succeed())

version.Next()  // invoke it every time the bundle is sent 
```

danmanor: Done

danmanor:
@yanmxa Seems like we still need the `sleep` - https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/stolostron_multicluster-global-hub/1123/pull-ci-stolostron-multicluster-global-hub-main-test-integration/1838145151910809600#1:build-log.txt%3A1806

test/integration/manager/status/security_alert_counts_handler_test.go
```diff
@@ -100,10 +100,13 @@ var _ = Describe("SecurityAlertCountsHandler", Ordered, func() {
...
 
-		version2 := eventversion.NewVersion()
-		version2.Incr()
+		By("Sync event1 with transport")
+		err := producer.SendEvent(ctx, *event1)
+		Expect(err).To(Succeed())
+
```
yanmxa, line 108: You need add the code `version.Next()  // invoke it every time the bundle is sent` to line 108

agent/pkg/status/controller/security/stackrox_syncer.go
```diff
@@ -459,6 +460,8 @@ func (s *StackRoxSyncer) produce(ctx context.Context, messageStruct any) error {
 		return fmt.Errorf("failed to send event %s: %v", *evt, err)
 	}
 
+	emitter.PostSend()
```
jhernand, line 463: If I understand correctly this call to `PostSend` is the one that will increment the version that we store in `s.version`. Is it safe to do this if we have multiple goroutines doing it simultaneously? It is unlikely, but if we have multiple instances of StackRox in the same hub then we could be syncing them simultaneously in different goroutines. Should we protect this with a mutex?

danmanor: Currently if we have multiple syncers so we will have multiple versions, So it shouldn't be a problem no ?

jhernand: If we had multiple syncers this wound't be a problem. But we have only one syncer for all the StackRox instances in the same hub.

danmanor: Oh so you mean theoretically we can have two go routines running `Sync` in the same time ?

jhernand: Yes, when there are multiple StackRox to sync we sync each of them in separate goroutines: https://github.com/stolostron/multicluster-global-hub/blob/e65419a39f9ccfc7ad56da714c3b228654d759cf/agent/pkg/status/controller/security/stackrox_syncer.go#L510-L522

test/integration/manager/status/security_alert_counts_handler_test.go
yanmxa: Yes, the `version.Next()` is invoked after sending the event bundle. the `version.Incr()` invoked when you want to send the event bundle.

yanmxa:
Thanks, @danmanor and @jhernand!

I see that the `stackrox_controller.go` includes additional logic, so we can retain the current implementation. However, if your goal is simply to send the resources to the global hub manager, you can directly leverage the following logic:

We've implemented a [GenericObjectSyncer](https://github.com/stolostron/multicluster-global-hub/blob/main/agent/pkg/status/controller/generic/generic_object_syncer.go), where you can simply provide the configuration for syncing the current object the global hub manager. For example, if you want to sync managed clusters, you can refer to the [managed cluster syncer](https://github.com/stolostron/multicluster-global-hub/blob/main/agent/pkg/status/controller/managedclusters/managed_cluster_syncer.go#L20-L45) for guidance.

Here’s how to configure it step by step:

1. **Configure how to watch the managed cluster resources**:
   ```go
   // controller config
   instance := func() client.Object { return &clusterv1.ManagedCluster{} }
   predicate := predicate.NewPredicateFuncs(func(object client.Object) bool { return true })
   ```

2. **Configure how to emit/send**:
   I added an annotation to each managed cluster to indicate which managed hub it belongs to:
   ```go
   // emitter config
   tweakFunc := func(object client.Object) {
      utils.MergeAnnotations(object, map[string]string{
         constants.ManagedClusterManagedByAnnotation: statusconfig.GetLeafHubName(),
      })
   }
   emitter := generic.ObjectEmitterWrapper(enum.ManagedClusterType, func(obj client.Object) bool {
      return true
   }, tweakFunc, false)
   ```

3. **Add the syncer to the manager**:
   ```go
   return generic.LaunchGenericObjectSyncer(
      "status.managed_cluster",
      mgr,
      generic.NewGenericController(instance, predicate),
      producer,
      statusconfig.GetManagerClusterDuration,
      []generic.ObjectEmitter{
         emitter,
      })
   ```
You can follow the above guidelines, it's internal implementation includes version control and multi-goroutine handling to avoid race conditions, among other features.

agent/pkg/status/controller/security/stackrox_syncer.go
```diff
@@ -139,6 +140,9 @@ func (b *StackRoxSyncerBuilder) Build() (result *StackRoxSyncer, err error) {
 		return
 	}
 
+	version := version.NewVersion()
+	version.Next()
```
yanmxa, line 144: The line 144 is not necessary. suggest remove the `version.Next()`. It invoked in the `emitter.PostSend()` every time you sent the event bundle

danmanor: My concern is about the first message sent. If we will not increase before the first message then lastVersion and current version are equal and `ShouldSend() == False`. Am I missing something ? @yanmxa

yanmxa: Get it. If you want to indicate that the bundle has been changed, use `version.Inrc()`(note: you can put it here or where the object is changed, that will be more reasonable). So it will pass `ShouldSend()`.

test/integration/manager/status/security_alert_counts_handler_test.go
danmanor: Done

agent/pkg/status/controller/security/stackrox_syncer.go
danmanor: Done

test/integration/manager/status/security_alert_counts_handler_test.go
```diff
@@ -42,10 +43,12 @@ var _ = Describe("SecurityAlertCountsHandler", Ordered, func() {
 			Source:    source1,
 		}
 		event := ToCloudEvent(leafHubName, string(enum.SecurityAlertCountsType), version, data)
+		version.Next()
```
yanmxa, line 46: Don't need the line `version.Next()`

test/integration/manager/status/security_alert_counts_handler_test.go
```diff
@@ -112,16 +117,12 @@ var _ = Describe("SecurityAlertCountsHandler", Ordered, func() {
...
 			Source:    source2,
 		}
-		event2 := ToCloudEvent(leafHubName, string(enum.SecurityAlertCountsType), version2, dataEvent2)
-
-		By("Sync events with transport")
-		err := producer.SendEvent(ctx, *event1)
-		Expect(err).To(Succeed())
-
-		time.Sleep(100 * time.Millisecond)
+		event2 := ToCloudEvent(leafHubName, string(enum.SecurityAlertCountsType), version, dataEvent2)
```
yanmxa, line 120: Need add the `version.Incr()` before sending the event2, you can add it between after line 119

test/integration/manager/status/security_alert_counts_handler_test.go
```diff
@@ -207,11 +206,12 @@ var _ = Describe("SecurityAlertCountsHandler", Ordered, func() {
 			DetailURL: DetailURL,
 			Source:    source1,
 		}
-		event1 := ToCloudEvent(leafHubName, string(enum.SecurityAlertCountsType), version1, dataEvent1)
+		event1 := ToCloudEvent(leafHubName, string(enum.SecurityAlertCountsType), version, dataEvent1)
```
yanmxa, line 209: Same as above, every time before sending an event, the `version.Incr()` must be invoked

test/integration/manager/status/security_alert_counts_handler_test.go
```diff
@@ -100,10 +101,15 @@ var _ = Describe("SecurityAlertCountsHandler", Ordered, func() {
...
+
+		By("Sync event1 with transport")
+		version.Incr()
```
yanmxa, line 107: Move this the line 107 before line 104, where the version is used

test/integration/manager/status/security_alert_counts_handler_test.go
```diff
@@ -44,8 +45,10 @@ var _ = Describe("SecurityAlertCountsHandler", Ordered, func() {
 		event := ToCloudEvent(leafHubName, string(enum.SecurityAlertCountsType), version, data)
 
 		By("Sync event with transport")
+		version.Incr()
```
yanmxa, line 48: Move the line `48` before line `45`, where it is used before sending

test/integration/manager/status/security_alert_counts_handler_test.go
```diff
@@ -112,16 +118,13 @@ var _ = Describe("SecurityAlertCountsHandler", Ordered, func() {
...
-
-		By("Sync events with transport")
-		err := producer.SendEvent(ctx, *event1)
-		Expect(err).To(Succeed())
-
-		time.Sleep(100 * time.Millisecond)
+		event2 := ToCloudEvent(leafHubName, string(enum.SecurityAlertCountsType), version, dataEvent2)
 
+		By("Sync event2 with transport")
+		version.Incr()
```
yanmxa, line 124: move the line `124` before line `121`

clyang82 (approved):

PR #1111, opened 2024-09-19
✨ support managedcluster migration event.

morvencao:
## Related issue(s)

Fixes #

## Tests
* [ ] Unit/function tests have been added and incorporated into `make unit-tests`.
* [x] Integration tests have been added and incorporated into `make integration-test`.
* [ ] E2E tests have been added and incorporated into `make e2e-test-all`.
* [ ] List other manual tests you have done.

PR #1129, opened 2024-09-25
enhance managedcluster migration controller and syncers.

morvencao:
## Related issue(s)

Follow up PR for #1111 
To address:
- cannot create msa when delete msa
- update the clusterrole and clusterrolebinding in the MigrationTo hub

## Tests

### DIFF
--- a/manager/pkg/migration/migration_controller.go
+++ b/manager/pkg/migration/migration_controller.go
@@ -23,6 +23,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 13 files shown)


manager/pkg/migration/migration_controller.go
```diff
@@ -100,6 +101,7 @@ func (m *MigrationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 					return false
 				},
 				DeleteFunc: func(e event.DeleteEvent) bool {
+					e.Object.SetDeletionTimestamp(&metav1.Time{Time: time.Now()})
```
clyang82, line 104: what does it mean?

morvencao: The controller runtime does not automatically set the deletion time when capturing a delete event; we must set it explicitly.

PR #1139, opened 2024-09-27
:sparkles: Add clusterInfo updating/deleting for inventory API

yanmxa:
## Related issue(s)

Fixes #

## Tests


agent/pkg/controllers/crd_controller.go
```diff
@@ -30,30 +30,29 @@ const (
...
+	restConfig      *rest.Config
+	agentConfig     *config.AgentConfig
+	transportClient transport.TransportClient
 }
 
 func (c *crdController) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
 	switch {
 	case request.Name == clusterManagersCRDName:
-		return c.reconcileClusterManagers(ctx, request)
+		return c.addACMController(ctx, request)
```
clyang82, line 43: it seems the Reconcile method can be called in the several times. it will start the ACMController and StackRoxCentrals with multiple instances.

clyang82: Let me fix it in a PR.

pkg/transport/controller/controller.go
```diff
@@ -18,37 +18,56 @@ import (
...
+	transportCallback TransportCallback
+	transportClient   *TransportClient
 
 	mutex sync.Mutex
 }
 
+type TransportClient struct {
+	consumer  transport.Consumer
+	producer  transport.Producer
+	inventory transport.InventoryClient
```
clyang82, line 46: if you defines the TransportClient, the inner objects should be `InventoryClient` and `KafkaClient`.

clyang82:
A General question:
How can we avoid send duplicated data to the inventory api? for example: if the agent is restarted, it always send the post request to the inventory-api

pkg/transport/controller/controller.go
yanmxa: @clyang82  I don't want to make the invoke flow deeper, the consumer and producer are for the event-based model(That means the producer/consumer is not only for Kafka; it can serve as other protocols), and the inventory client is for the request-response model. 

I can rename the inventory client with `RestfulClient` to align with it. What do you think of it?

yanmxa:
> A General question: How can we avoid send duplicated data to the inventory api? for example: if the agent is restarted, it always send the post request to the inventory-api

let it leverage the time filter to deduplicate the message?

agent/pkg/controllers/crd_controller.go
clyang82: it is my fault. we check the CRD name to start the controller. it won't start multiple instances.

pkg/transport/controller/controller.go
clyang82: Let us use `restfulClient`

yanmxa:
The left uncovered code is the runtime `addManager`. Let's skip them!

/override ci/prow/sonarcloud

clyang82 (approved):
LGTM

PR #1148, opened 2024-09-29
fix policy number is not right

ldpliu:
## Summary
In code logic, if the policy and hub info in synced to db, but do not synced heart beat. then we detached the hub cluster. The related policy and hubinfo will not be removed forever. So in this pr, we send the event immediately instead of waiting interval.

Will enhance this in future: https://issues.redhat.com/browse/ACM-14560

## Related issue(s)
https://issues.redhat.com/browse/ACM-14401
Fixes #

## Tests


yanmxa:
💯 Good Catch!

PR #1250, opened 2024-11-29
:sparkles: Use Kafka KRaft

clyang82:
## Summary

Use KRaft mode since it is GAed. besides it, also address the following issues in this PR:
1. remove plain listener for Kafka to disable anonymous access
2. 

## Related issue(s)

Fixes # https://issues.redhat.com/browse/ACM-15625

## Tests


samples/config/confluent_config.go
```diff
@@ -219,8 +219,8 @@ func GetConfluentConfigMapByUser(c client.Client, namespace, clusterName, userNa
 	cm := config.GetBasicConfigMap()
 	for _, condition := range kafkaCluster.Status.Conditions {
 		if *condition.Type == "Ready" && *condition.Status == "True" {
-			clusterCaCert := kafkaCluster.Status.Listeners[1].Certificates[0]
-			_ = cm.SetKey("bootstrap.servers", *kafkaCluster.Status.Listeners[1].BootstrapServers)
+			clusterCaCert := kafkaCluster.Status.Listeners[0].Certificates[0]
```
yanmxa, line 222: There are two listeners for the kafka connection. the first one is internal, and the other is external. You mean only keep one type?

ldpliu:
Did you test if the upgrade works well?

samples/config/confluent_config.go
clyang82: Yes. remove plain listener for Kafka to disable anonymous access.

clyang82:
> Did you test if the upgrade works well?

will handle the upgrade in a separate user story.

yanmxa:
It's a significant improvement in transport optimization, especially in its footprint!

/lgtm

PR #1252, opened 2024-11-29
fix manager do not recreted after delete manager deploy

ldpliu:
## Summary
When operator restart, the transport connection is null, so the controllers(manager) which depend on the transport connection will not be start. But when transport connection is set, the kafka component status in mgh is not updated, so controller-runtime will not trigger reconcile to start manager controller. 
In this pr, we update the kafka component status in mgh when transport connection set.
The storage connection has the same problem, so we do it as the same way.
## Related issue(s)

Fixes #

## Tests


clyang82 (changes requested):

clyang82 (approved):
LGTM

PR #1272, opened 2024-12-12
fix dashboard - postgres database has no data

ldpliu:
## Summary
The postgres url is not right and sshmode should be disable
## Related issue(s)
https://issues.redhat.com/browse/ACM-16333
Fixes #
image

## Tests

### DIFF
--- a/operator/pkg/controllers/storage/postgres_statefulset.go
+++ b/operator/pkg/controllers/storage/postgres_statefulset.go
@@ -3,6 +3,7 @@ package storage
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


operator/pkg/controllers/storage/postgres_statefulset.go
```diff
@@ -99,7 +100,9 @@ func InitPostgresByStatefulset(ctx context.Context, mgh *globalhubv1alpha4.Multi
 				PostgresReadonlyUsername:     credential.postgresReadonlyUsername,
 				PostgresReadonlyUserPassword: credential.postgresReadonlyUserPassword,
 				StorageClass:                 mgh.Spec.DataLayerSpec.StorageClass,
-				PostgresURI:                  builtinPartialPostgresURI,
+				// Postgres exporter should disable sslmode
+				// https://github.com/prometheus-community/postgres_exporter?tab=readme-ov-file#environment-variables
+				PostgresURI: strings.ReplaceAll(builtinPartialPostgresURI, "sslmode=verify-ca", "sslmode=disable"),
```
yanmxa, line 105: Can we change it to Grafana Postgres URL if it is only for the dashboard?

operator/pkg/controllers/storage/postgres_statefulset.go
```diff
@@ -28,7 +29,7 @@ var (
 	builtinPostgresCertName   = fmt.Sprintf("%s-cert", BuiltinPostgresName)
 	builtinPostgresConfigName = fmt.Sprintf("%s-config", BuiltinPostgresName)
 	builtinPostgresInitName   = fmt.Sprintf("%s-init", BuiltinPostgresName)
-	builtinPartialPostgresURI = fmt.Sprintf("@%s.%s.svc:5432/hoh?sslmode=verify-ca", BuiltinPostgresName,
```
yanmxa, line 31: I see you have added the '@' character in the following. Why remove it from here? Do it encounter some special case with it?

PR #1318, opened 2025-01-08
:sparkles: Support SQL init from API

yanmxa:
### Option 1: execute the SQL directly
The database cannot be created within a SQL block as it needs to be created separately. For instance, if you execute the following SQL:

```sql
-- Create the test database if it doesn't exist
CREATE DATABASE test;

-- Create the COMPANY table in the 'test' database
CREATE TABLE IF NOT EXISTS public.COMPANY (
  ID INT PRIMARY KEY NOT NULL,
  NAME TEXT NOT NULL,
  AGE INT NOT NULL,
  ADDRESS CHAR(50),
  SALARY REAL
);
```

It will result in the error:
`2025-01-08 *** ERROR: CREATE DATABASE cannot run inside a transaction block.`

To resolve this, the database must be created separately, outside of any transaction block. Only after the database is created can you proceed to create tables or other objects within it.

### Option 2: use the initdb 

The entrypoint script checks if the database has already been initialized by looking for the presence of files in the PostgreSQL data directory (/var/lib/postgresql/data). If these files exist, it skips running initialization scripts again.

References: https://www.postgresql.org/docs/16/creating-cluster.html

The Postgres-start will execute every time the pods restart: https://github.com/sclorg/postgresql-container/tree/master/16#postgresql-start 

```bash
pg_ctl: another server might be running; trying to start server anyway
waiting for server to start....2025-01-09 04:55:37.590 UTC [24] LOG:  redirecting log output to logging collector process
2025-01-09 04:55:37.590 UTC [24] HINT:  Future log output will appear in directory "log".
 done
server started
/var/run/postgresql:5432 - accepting connections
=> sourcing /opt/app-root/src/postgresql-start/postgresql-start.sh ...
ERROR:  database "inventory" already exists
```

## Related issue(s)

Fixes # https://issues.redhat.com/browse/ACM-16818

## Tests


yanmxa:
close this pr in favor of https://github.com/stolostron/multicluster-global-hub/pull/1322

PR #1319, opened 2025-01-08
:sparkles: Install GH into a brownfield

clyang82:
## Summary

InstallAgentOnHub determines whether deploy the Global Hub Agent on hub cluster.
If set to true, the Global Hub Agent will be installed on the hub cluster only.
If set to false, the Global Hub Agent will not be installed on the hub cluster.
Currently, switching the value of this field is not supported after the Global Hub is installed.

## Related issue(s)

Fixes # https://issues.redhat.com/browse/ACM-15099

## Tests
- tested deploy the agent into local-cluster
- tested deploy the agent into hub cluster which has local-cluster="true" label

### DIFF
--- a/operator/api/operator/v1alpha4/multiclusterglobalhub_types.go
+++ b/operator/api/operator/v1alpha4/multiclusterglobalhub_types.go
@@ -107,13 +107,13 @@ type MulticlusterGlobalHubSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
 	// +optional
 	EnableMetrics bool `json:"enableMetrics"`
-	// InstallAgentOnHub determines whether deploy the Global Hub Agent on hub cluster.
-	// If set to true, the Global Hub Agent will be installed on the hub cluster only.
-	// If set to false, the Global Hub Agent will not be installed on the hub cluster.
+	// InstallAgentOnLocal determines whether deploy the Global Hub Agent on the local hub cluster or not.
+	// If set to true, the Global Hub Agent will be installed on the local hub cluster only.
+	// If set to false, the Global Hub Agent will not be installed on the local hub cluster.
 	// Currently, switching the value of this field is not supported after the Global Hub is installed.
 	// +kubebuilder:default=false
 	// +optional
-	InstallAgentOnHub bool `json:"installAgentOnHub,omitempty"`
+	InstallAgentOnLocal bool `json:"InstallAgentOnLocal,omitempty"`
 }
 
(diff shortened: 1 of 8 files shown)


operator/api/operator/v1alpha4/multiclusterglobalhub_types.go
```diff
@@ -107,6 +107,13 @@ type MulticlusterGlobalHubSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
 	// +optional
 	EnableMetrics bool `json:"enableMetrics"`
+	// InstallAgentOnHub determines whether deploy the Global Hub Agent on hub cluster.
```
yanmxa, line 110: Since the global hub involves different hubs - global hub, managed hub.

"on hub cluster" -> "on global hub cluster"?

operator/api/operator/v1alpha4/multiclusterglobalhub_types.go
```diff
@@ -107,6 +107,13 @@ type MulticlusterGlobalHubSpec struct {
...
+	// If set to true, the Global Hub Agent will be installed on the hub cluster only.
+	// If set to false, the Global Hub Agent will not be installed on the hub cluster.
+	// Currently, switching the value of this field is not supported after the Global Hub is installed.
```
yanmxa, line 113: That means that value should be set when creating the instance. After creating, it won't support being updated. 
If so, we should clarify the value is immutable and add validation for this logic

yanmxa: If there is no validation here, the desired status and actual status will be consistent after the user updates it

clyang82: This is just an initial version. Based on my understanding, we do not support switching the value right now. but we may support later.

yanmxa: Okay, Could you log a Jira issue(or GitHub issue), I'm concerned we might overlook it in the future.

operator/api/operator/v1alpha4/multiclusterglobalhub_types.go
clyang82: rename to `InstallAgentOnLocal`

operator/api/operator/v1alpha4/multiclusterglobalhub_types.go
clyang82: https://github.com/stolostron/multicluster-global-hub/issues/1321 log this github issue to trace. Thanks.

PR #1332, opened 2025-01-16
:sparkles: generate kafka config script

clyang82:
## Related issue(s)

Fixes #

## Tests

### DIFF
--- a/tools/generate-kafka-config.sh
+++ b/tools/generate-kafka-config.sh
@@ -1,0 +1,124 @@
+#!/bin/bash
+
+global_hub_namespace=${2:-"multicluster-global-hub"}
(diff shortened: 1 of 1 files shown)


tools/generate-kafka-config.sh
```diff
@@ -0,0 +1,124 @@
...
+  echo "kafka.yaml Content:"
+  echo "==========================================="
+  echo ""
+  cat kafka.yaml
+  echo ""
+}
+
+# create a KafkaUser and KafkaTopic for the given name
+createKafkaTopic $1
+if [ $? -ne 0 ]; then
```
yanmxa, line 112: we can `set -e` to avoid checking the non-zero every time

PR #1425, opened 2025-02-26
:memo: Limit the kafka memory with JVM configuration

yanmxa:
## Related issue(s)

Fixes # https://issues.redhat.com/browse/ACM-18122

## Tests


clyang82 (commented):
Do we have some errors in kafka logs during your tests?
Do we have data process timing comparison when set to 1Gi?

yanmxa:
1. I haven't found any data loss, and there are no error messages in the Kafka logs
2. I haven't compared the timing consumption, as when I import the hubs into the global hub, it syncs the clusters and policies without much delay, as shown in the following image. Therefore, I believe the time required at this scale is acceptable.

image

PR #1443, opened 2025-03-04
fix agent can not connect to kafka in dr case

ldpliu:
## Related issue(s)
https://issues.redhat.com/browse/ACM-18120
Fixes #

## Tests

### DIFF
--- a/doc/reserved_names/README.md
+++ b/doc/reserved_names/README.md
@@ -22,7 +22,7 @@ List all annotations are used by multicluster global hub.
 |global-hub.open-cluster-management.io/import-cluster-in-hosted=true\|false | This annotation is used to identify if managedhub cluster should be imported in hosted mode |
 | global-hub.open-cluster-management.io/with-inventory                | This annotation is used to identify the common inventory is deployed.                                                                  |
 | global-hub.open-cluster-management.io/with-stackrox-integration | This annotation enables the experimental integration with [Stackrox](https://github.com/stackrox).|
-| global-hub.open-cluster-management.io/resign-kafka-client-secret | This annotation is used to identify if the kafka client secret is resynced in agent.|
+| global-hub.open-cluster-management.io/kafka-cluster-id | This annotation save the kafka cluster id in transport-config secret in agent part, it is used to identify if the kafka cluster changed.|
 
 # Finalizer
 
--- a/pkg/transport/controller/controller.go
+++ b/pkg/transport/controller/controller.go
@@ -244,41 +244,51 @@ func (c *TransportCtrl) ReconcileKafkaCredential(ctx context.Context, secret *co
 	return true, nil
 }
 
-// Resync the kafka client secret because we recreate the kafka cluster.
-// TODO: Remove the code in globalhub 1.5
+// Resync the kafka client secret because we recreate the kafka cluster in globalhub 1.4 and restore case.
 func (c *TransportCtrl) ResyncKafkaClientSecret(ctx context.Context, kafkaConn *transport.KafkaConfig, secret *corev1.Secret) error {
-	log.Debugf("resync kafka client secret: %v", kafkaConn.IsNewKafkaCluster)
-	if !kafkaConn.IsNewKafkaCluster {
+	if kafkaConn.ClusterID == "" {
 		return nil
 	}
 
+	isKafkaClusterIdEqual := true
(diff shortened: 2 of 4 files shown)


doc/reserved_names/README.md
```diff
@@ -22,7 +22,7 @@ List all annotations are used by multicluster global hub.
 |global-hub.open-cluster-management.io/import-cluster-in-hosted=true\|false | This annotation is used to identify if managedhub cluster should be imported in hosted mode |
 | global-hub.open-cluster-management.io/with-inventory                | This annotation is used to identify the common inventory is deployed.                                                                  |
 | global-hub.open-cluster-management.io/with-stackrox-integration | This annotation enables the experimental integration with [Stackrox](https://github.com/stackrox).|
-| global-hub.open-cluster-management.io/resign-kafka-client-secret | This annotation is used to identify if the kafka client secret is resynced in agent.|
+| global-hub.open-cluster-management.io/kafka-cluster-id | This annotation save the kafka cluster id, it is used to identify if the kafka changed in agent.|
```
yanmxa, line 25: We need to clarify which object the annotation will be added to. Over time, we might lose track of the information

pkg/transport/controller/controller.go
```diff
@@ -244,44 +245,65 @@ func (c *TransportCtrl) ReconcileKafkaCredential(ctx context.Context, secret *co
...
 	}
 	log.Infof("update transport config secret")
 
-	transportSecret.Annotations[constants.ResyncKafkaClientSecretAnnotation] = "true"
+	transportSecret.Annotations[constants.KafkaClusterIdAnnotation] = sanitizeClusterID(kafkaConn.ClusterID)
 	return c.runtimeClient.Update(ctx, transportSecret)
 }
 
+// sanitizeClusterID replace the illegal character for label
+func sanitizeClusterID(clusterID string) string {
```
yanmxa, line 297: Why do we need to sanitize the clusterID, and that will lose the original information

pkg/transport/controller/controller.go
```diff
@@ -244,44 +245,65 @@ func (c *TransportCtrl) ReconcileKafkaCredential(ctx context.Context, secret *co
...
 	}
+
 	transportSecret := secret.DeepCopy()
 	if transportSecret.Annotations == nil {
 		transportSecret.Annotations = make(map[string]string)
 	}
 	log.Infof("update transport config secret")
 
-	transportSecret.Annotations[constants.ResyncKafkaClientSecretAnnotation] = "true"
+	transportSecret.Annotations[constants.KafkaClusterIdAnnotation] = sanitizeClusterID(kafkaConn.ClusterID)
```
yanmxa, line 292: The `kafkaClusterID` originally comes from the payload of the transportSecret, which is generated by the gh manifestwork. However, the annotation is added by the global hub agent.
Wouldn't adding this information risk being overridden? I noticed we've done this before - just curious :)

ldpliu: addon framework will merge the existing annotation

pkg/transport/controller/controller.go
ldpliu: > Why do we need to sanitize the clusterID, and that will lose the original information

The cluster is an uuid which is generated based on base64 encoding. It may contain the `/` or '+'. But the label value do not allow that.

doc/reserved_names/README.md
ldpliu: Done

pkg/transport/controller/controller.go
ldpliu: Removed the logic

yanmxa:
The transport controller is intuitively designed to detect changes in the transport credentials(include the signed certificate), and then trigger a reconnection to the Kafka cluster when necessary.


For the current case/issue:

The first priority in resolving this issue is to ensure that the addon renews its signed certificate when it expires or becomes invalid, allowing the Kafka cluster to reconnect automatically. I have opened an issue in the addon-framework to track this: [#304](https://github.com/open-cluster-management-io/addon-framework/issues/304)

However, since the addon registration may not currently support certificate renewal, this solution looks good to me for the current GH case.

/cc @ldpliu @clyang82

ldpliu:
I tested the whole flow(backup and restore), and it works fine.

clyang82 (approved):
LGTM

PR #1510, opened 2025-04-02
:seedling: Run on monday

clyang82:
## Summary

e2e is not stable recently since we added QE test cases. we do not have time to resolve the issues. but we need to keep tracking the failure cases to find out the root case in the future. will track the failure cases in a github issue.
## Related issue(s)

Fixes #

## Tests


clyang82:
in favor of https://github.com/stolostron/multicluster-global-hub/pull/1532

PR #1536, opened 2025-04-14
:sparkles: Migrate managed clusters

clyang82:
## Summary
include `gh-migration` to migrate the resource from the source clusters to the target cluster. we cannot use the existing topics, because the permission is not proper set. In this case, we want to send the resources from the source clusters and receive from the target cluster directly. Keep in mind, try to relieve the dependence of db.

## Related issue(s)

Fixes # [ACM-19539](https://issues.redhat.com/browse/ACM-19539)

## Tests
1. create a managed cluster with foo=bar label and foo=bar annotation in local-cluster
2. migrate this managed cluster from local-cluster to new-hub cluster.
3. check the foo=bar label and annotation are there in new-hub cluster

### DIFF
--- a/agent/pkg/controllers/init_controller.go
+++ b/agent/pkg/controllers/init_controller.go
@@ -67,6 +67,11 @@ func (c *initController) addACMController(ctx context.Context, request ctrl.Requ
 		return ctrl.Result{}, fmt.Errorf("failed to add the syncer: %w", err)
 	}
 
+	// add spec controllers
+	if err := agentspec.AddToManager(ctx, c.mgr, c.transportClient, c.agentConfig); err != nil {
+		return ctrl.Result{RequeueAfter: 10 * time.Second}, fmt.Errorf("failed to add spec syncer: %w", err)
+	}
+
 	// only enable the status controller in the standalone mode
 	if c.agentConfig.Standalone {
 		return ctrl.Result{}, nil
@@ -80,11 +85,6 @@ func (c *initController) addACMController(ctx context.Context, request ctrl.Requ
 	// Need this controller to update the value of clusterclaim version.open-cluster-management.io
 	if err := AddVersionClusterClaimController(c.mgr); err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to add controllers: %w", err)
-	}
-
-	// add spec controllers
-	if err := agentspec.AddToManager(ctx, c.mgr, c.transportClient, c.agentConfig); err != nil {
-		return ctrl.Result{RequeueAfter: 10 * time.Second}, fmt.Errorf("failed to add spec syncer: %w", err)
 	}
 
 	// all the controller started, then add the lease controller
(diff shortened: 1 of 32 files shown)


agent/pkg/controllers/init_controller.go
```diff
@@ -67,6 +67,11 @@ func (c *initController) addACMController(ctx context.Context, request ctrl.Requ
 		return ctrl.Result{}, fmt.Errorf("failed to add the syncer: %w", err)
 	}
 
+	// add spec controllers
```
yanmxa, line 70: Why enable the specPath even if the standalone agent is enable?

agent/pkg/spec/syncers/migration_to_syncer.go
```diff
@@ -109,6 +126,93 @@ func (s *managedClusterMigrationToSyncer) Sync(ctx context.Context, payload []by
...
 
+func (s *managedClusterMigrationToSyncer) StartMigrationConsumer(ctx context.Context) error {
+	// initialize the gh-migration consumer
+	if s.migrationConsumer == nil && s.transportConfig != nil {
+		var migrationCtx context.Context
+		migrationCtx, s.migrationConsumerCtxCancel = context.WithCancel(ctx)
+
+		var err error
+		s.log.Infof("start the kafka consumer to consume the migration topic")
+		s.migrationConsumer, err = consumer.NewGenericConsumer(s.transportConfig,
```
yanmxa, line 137: We might need to explicitly specify the consumer fetch message from the earliest. In case of consuming the message from the Latest. That means it might lose the data before it starts

agent/pkg/spec/syncers/migration_from_syncer.go
```diff
@@ -42,35 +43,56 @@ const (
...
+
 func (s *managedClusterMigrationFromSyncer) Sync(ctx context.Context, payload []byte) error {
 	// handle migration.from cloud event
 	managedClusterMigrationEvent := &bundleevent.ManagedClusterMigrationFromEvent{}
 	if err := json.Unmarshal(payload, managedClusterMigrationEvent); err != nil {
 		return err
 	}
 	s.log.Debugf("received managed cluster migration event %s", string(payload))
+	// initialize the gh-migration producer
+	if s.migrationProducer == nil && s.transportConfig != nil {
```
yanmxa, line 80: It seems we can't get the confirmation message from the destination hub. Maybe we need a mechanism to ensure the message is received by the destination hub cluster?

agent/pkg/controllers/init_controller.go
clyang82: need to use spec to receive the migration event from migration controller.

agent/pkg/spec/syncers/migration_to_syncer.go
clyang82: we are using `earliest` by default. https://github.com/stolostron/multicluster-global-hub/blob/d823a506825e405c56c88c1af7c60b00b579ef51/pkg/transport/config/confluent_config.go#L44

agent/pkg/spec/syncers/migration_from_syncer.go
clyang82: if only the message is sent successfully by the producer, the consumer in the target hub cluster should be able to receive the message. just like other controllers send the message to spec/status topics.  we need to send the confirmation message to the migration controller to update the status. need a separate PR for it.

yanmxa: OK

PR #1545, opened 2025-04-17
:sparkles: Add cluster validation before starting migration process

yanmxa:
## Related issue(s)

Fixes # https://issues.redhat.com/browse/ACM-19782

## Tests

### DIFF
--- a/manager/pkg/migration/migration_cleaning.go
+++ b/manager/pkg/migration/migration_cleaning.go
@@ -23,7 +23,8 @@ func (m *ClusterMigrationController) completed(ctx context.Context,
 		return false, nil
 	}
 
-	if meta.IsStatusConditionTrue(mcm.Status.Conditions, migrationv1alpha1.ConditionTypeCleaned) {
+	if meta.IsStatusConditionTrue(mcm.Status.Conditions, migrationv1alpha1.ConditionTypeCleaned) ||
+		mcm.Status.Phase != migrationv1alpha1.PhaseCleaning {
 		return false, nil
 	}
 
--- a/manager/pkg/migration/migration_validating.go
+++ b/manager/pkg/migration/migration_validating.go
@@ -13,14 +13,15 @@ import (
 
 const (
 	ConditionReasonClusterNotFound   = "ClusterNotFound"
+	ConditionReasonClusterConflict   = "ClusterConflict"
 	ConditionReasonResourceValidated = "ResourceValidated"
 )
 
 // Validation:
 // 1. Verify if the migrating clusters exist in the current hubs
 //   - If the clusters do not exist in both the from and to hubs, report a validating error and mark status as failed
-//   - If the clusters exist in the destination hub, mark the status as completed.
 //   - If the clusters exist in the source hub, mark the status as validated and change the phase to initializing.
+//   - If the clusters exist in the destination hub, mark the status as failed, raise the 'ClusterConflict' message.
 func (m *ClusterMigrationController) validating(ctx context.Context,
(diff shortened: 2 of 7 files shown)


manager/pkg/migration/migration_validating.go
```diff
@@ -0,0 +1,119 @@
...
+	mcm *migrationv1alpha1.ManagedClusterMigration,
+) (bool, error) {
+	if !mcm.DeletionTimestamp.IsZero() {
+		return false, nil
+	}
+
+	if mcm.Status.Phase == "" {
+		mcm.Status.Phase = migrationv1alpha1.PhaseValidating
+		if err := m.Client.Status().Update(ctx, mcm); err != nil {
+			return false, err
```
ldpliu, line 34: What's the validating mean? why you return false here?

manager/pkg/migration/migration_validating.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	// not found validated clusters
+	if len(notFoundClusters) > 0 {
+		mcm.Status.Phase = migrationv1alpha1.PhaseFailed
+		condStatus = metav1.ConditionFalse
+		condReason = ConditionReasonClusterNotFound
+		clusters := notFoundClusters
+		if len(notFoundClusters) > 3 {
+			clusters = append(clusters[:3], "...")
```
ldpliu, line 90: what's this mean?

manager/pkg/migration/migration_validating.go
yanmxa: It's an error `!= nil`, it will requeue by default

It is used to verify MCM is validated. For now, It will assert whether the migrated clusters exists, if not, report it in the cr status.

manager/pkg/migration/migration_validating.go
```diff
@@ -0,0 +1,119 @@
...
+
+const (
+	ConditionReasonClusterNotFound   = "ClusterNotFound"
+	ConditionReasonResourceValidated = "ResourceValidated"
+)
+
+// Validation:
+// 1. Verify if the migrating clusters exist in the current hubs
+//   - If the clusters do not exist in both the from and to hubs, report a validating error and mark status as failed
+//   - If the clusters exist in the destination hub, mark the status as completed.
```
clyang82, line 22: why do we need to care the clusters in the target hub cluster?

manager/pkg/migration/migration_cleaning.go
```diff
@@ -23,7 +23,8 @@ func (m *ClusterMigrationController) completed(ctx context.Context,
 		return false, nil
 	}
 
-	if meta.IsStatusConditionTrue(mcm.Status.Conditions, migrationv1alpha1.ConditionTypeCleaned) {
+	if meta.IsStatusConditionTrue(mcm.Status.Conditions, migrationv1alpha1.ConditionTypeCleaned) ||
+		mcm.Status.Phase != migrationv1alpha1.PhaseCleaning {
```
clyang82, line 27: could you add some comments? why not equal to `PhaseCleaning`?

manager/pkg/migration/migration_validating.go
yanmxa: If the migrating clusters are already in the target hub cluster, we don't need to start the following step, like initializing, ..., just mark it as completed

manager/pkg/migration/migration_cleaning.go
yanmxa: It skips the following validation logic if it equals `PhaseCleaning `.

PR #1651, opened 2025-05-13
fix rename local-cluster issue

ldpliu:
## Related issue(s)

Fixes #
https://issues.redhat.com/browse/ACM-20560
## Tests

### DIFF
--- a/agent/cmd/main.go
+++ b/agent/cmd/main.go
@@ -88,8 +88,8 @@ func doMain(ctx context.Context, agentConfig *configs.AgentConfig, restConfig *r
 		return fmt.Errorf("failed to create manager: %w", err)
 	}
 	transportSecretName := constants.GHTransportConfigSecret
-	if agentConfig.LeafHubName == constants.LocalClusterName {
-		transportSecretName = constants.GHTransportConfigSecret + "-" + constants.LocalClusterName
+	if agentConfig.TransportConfigSecretName != "" {
+		transportSecretName = agentConfig.TransportConfigSecretName
 	}
 	// add transport ctrl to manager, also load the transportConfig(from secret) into the agentConfig
 	err = controller.NewTransportCtrl(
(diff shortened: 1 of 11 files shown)


operator/pkg/controllers/agent/local_agent_controller.go
```diff
@@ -150,15 +204,61 @@ func (s *LocalAgentController) Reconcile(ctx context.Context, req ctrl.Request)
 		mgh.Spec.Tolerations,
 		mgh,
 		clusterName,
+		getTransportSecretName(),
 	)
 }
 
+// getCurrentClusterName returns the current cluster name from the agent deployment
+func getCurrentClusterName(ctx context.Context, c client.Client, namespace string) (error, string) {
```
clyang82, line 212: the current cluster can be from agentConfig.LeafHubName. it is unnecessary to get it from deployment.

agent/cmd/main.go
```diff
@@ -120,6 +120,8 @@ func parseFlags() *configs.AgentConfig {
 
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 	pflag.StringVar(&agentConfig.LeafHubName, "leaf-hub-name", "", "The name of the leaf hub.")
+	pflag.StringVar(&agentConfig.TransportConfigSecretName, "transport-config-secret",
```
clyang82, line 123: why do you need to introduce this new param?

operator/pkg/controllers/agent/local_agent_controller.go
ldpliu: > the current cluster can be from agentConfig.LeafHubName. it is unnecessary to get it from deployment.

This func works in operator, not in agent. So there is no agentConfig

agent/cmd/main.go
ldpliu: > why do you need to introduce this new param?

For local agent, the transport config secret name is `transport-config-<Local-Cluster-Name>`, and we can not known if the current agent is local agent, so we should use a new flag to identify the transport secret name.

clyang82 (approved):
LGTM

PR #1685, opened 2025-05-21
get next migration when it's failed

ldpliu:
## Related issue(s)

Fixes #

## Tests


manager/pkg/migration/migration_controller.go
```diff
@@ -217,9 +216,9 @@ func (m *ClusterMigrationController) getCurrentMigration(ctx context.Context,
 		if migration.Status.Phase == migrationv1alpha1.PhaseCompleted {
 			continue
 		}
-		// skip the migration which is failed and cleaned
-		if migration.Status.Phase == migrationv1alpha1.PhaseFailed &&
-			meta.IsStatusConditionTrue(migration.Status.Conditions, migrationv1alpha1.ConditionTypeCleaned) {
+		// skip the migration which is failed
+		// (we do not care the clean process, the clean process should not impact current migration)
```
clyang82, line 220: we need to care about the clean process. i.e.: if you migrate the managed clusters from the same source cluster and the same target cluster, the cleanup process can be able to clean the resources (bootstrap kubeconfig, etc) you have created for the new migration process.

ldpliu: @yanmxa @yahliu

clyang82: after offline discussed, we agreed that here just check if we have condition of cleaned or not. only add the ConditionTypeCleaned condition if reported from agent.

ldpliu: Done

clyang82: >  only add the ConditionTypeCleaned condition if reported from agent.
Don't need to add this logic?

PR #1689, opened 2025-05-22
:sparkles:  enable the global hub addon in hosted mode - renaming

yanmxa:
## Related issue(s)

Reference: https://github.com/stolostron/multicluster-global-hub/pull/1686

## Tests


agent/pkg/spec/syncers/migration_to_syncer.go
```diff
@@ -90,33 +92,24 @@ func (s *migrationTargetSyncer) Sync(ctx context.Context, evt *cloudevents.Event
...
-						if e := s.registering(ctx, managedClusterMigrationToEvent, notAvailableManagedClusters); e != nil {
-							log.Infof("waiting the migrating clusters are available: %v", e)
-							return false, nil
-						}
-						return true, nil
-					})
+				err := s.registering(ctx, managedClusterMigrationToEvent)
 				if err != nil {
-					reportErrMessage = fmt.Sprintf("failed to register these clusters [%s]: %v",
-						strings.Join(notAvailableManagedClusters, ", "), err)
```
clyang82, line 106: why do you want to remove it? it is designed to include the unavailable clusters in the msg. do we have a bug for it?

operator/pkg/config/addon_config.go
```diff
@@ -34,7 +34,7 @@ import (
 
 const SignerName = "open-cluster-management.io/globalhub-signer"
 
-var isGlobalhubAgentRemoved = true
```
clyang82, line 37: why rename to AddonResource?

operator/pkg/controllers/agent/addon/managedclusteraddon_controller.go
clyang82, line 1: it is not only handle managedclusteraddon, but also handle transportresource. maybe keep as it is.

agent/pkg/spec/syncers/migration_to_syncer.go
```diff
@@ -90,33 +92,24 @@ func (s *migrationTargetSyncer) Sync(ctx context.Context, evt *cloudevents.Event
 
 		if managedClusterMigrationToEvent.Stage == migrationv1alpha1.PhaseRegistering {
 			go func() {
-				log.Infof("registering managed cluster migration")
+				log.Info("registering managed cluster migration")
 				reportErrMessage := ""
-				notAvailableManagedClusters := []string{}
-				err := wait.PollUntilContextTimeout(ctx, 10*time.Second, 10*time.Minute, true,
```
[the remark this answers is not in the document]
yanmxa: We don't need wait every error message when registering. For some error reason like "no cluster in the received event", just report it directly.

So I just move it internal to where it need to PollUntil.

agent/pkg/spec/syncers/migration_to_syncer.go
yanmxa: We should report why the cluster is not registered, not the message like this "failed to register these clusters: ... time out"

operator/pkg/config/addon_config.go
yanmxa: This is align with the `ManagedClusterAddonController`. 

The `Global Hub Agent` is a big scope. `local_agent`, `hosted_agent`, or `default_agent`?

operator/pkg/controllers/agent/addon/managedclusteraddon_controller.go
yanmxa: The mainly resources is `managedclusteraddon`, all the other resources is based on that. the original name is `default_agent_controller`, but the internal is handle both default mode agent and hosted mode agent. The only thing we can extract from the common logic is: **they all based on the managedclusteraddon to create its related resources(like kakfa user and topic)**

Also rename it can distinguish it with the `agent_addon.go`.

PR #1816, opened 2025-07-10
check isResourceremoved before prune the resources in each controllers

ldpliu:
## Related issue(s)
https://issues.redhat.com/browse/ACM-19702
Fixes #

## Tests


ldpliu:
use ai to fix the code issue in https://github.com/stolostron/multicluster-global-hub/pull/1816/files#diff-ff8fd2296cfd658ed6708d8e58d8e3ac804532b5ae0917d43d44492606b27abbR552

operator/pkg/controllers/agent/local_agent_controller.go
```diff
@@ -138,14 +138,12 @@ func (s *LocalAgentController) Reconcile(ctx context.Context, req ctrl.Request)
...
-		}
-		err := pruneAgentResources(ctx, s.GetClient(), mgh.Namespace)
+		log.Debugf("deleting mgh in local agent controller")
+		err = utils.HandleMghDelete(ctx, &isResourceRemoved, mgh.Namespace, s.pruneAgentResources)
 		if err != nil {
 			return ctrl.Result{}, err
 		}
-		isResourceRemoved = true
+		log.Debugf("deleted local agent resources, isResourceRemoved:%v", isResourceRemoved)
 		return ctrl.Result{}, nil
```
yanmxa, line 147: Seems can concise it into 

return ctrl.Result{}, err

yanmxa:
Hi @ldpliu,

Can you clarify why do we need to do this? I see you have set the flag to assert where the resource is removed before this modification

ldpliu:
> Hi @ldpliu,
> 
> Can you clarify why do we need to do this? I see you have set the flag to assert where the resource is removed before this modification

This pr handle the case that for each controller, if the resource already removed, it should not run the pure logic any more.

PR #1867, opened 2025-08-05
Validate mcm clusters

ldpliu:
## Related issue(s)

Fixes #

## Tests


yanmxa:
```
Events:
  Type     Reason            Age                    From                  Message
  ----     ------            ----                   ----                  -------
  Warning  ValidationFailed  3m27s (x2 over 3m27s)  migration-controller  cluster cluster1 not found in database;
  Warning  ValidationFailed  3m27s (x2 over 3m27s)  migration-controller  cluster mc is already on hub mh1;
  Warning  ValidationFailed  3m27s (x2 over 3m27s)  migration-controller  cluster mc1 not found in database;
  Warning  ValidationFailed  3m27s (x2 over 3m27s)  migration-controller  cluster mc2 not found in databa
```
Is it possible to change the Warning type into Erroring?

manager/pkg/migration/migration_validating.go
```diff
@@ -169,33 +166,173 @@ func IsValidResource(resource string) error {
...
+
+	if len(failedClusters) > 0 {
+		// Set the condition reason based on the failure type
+		condition.Reason = conditionReason
+		return fmt.Errorf(validationFailureFormat, messageClusters(failedClusters))
 	}
-	return messages
+
+	// Record successful validation event
+	m.EventRecorder.Eventf(mcm, corev1.EventTypeNormal, "ValidationSucceeded",
```
yanmxa, line 283: A question: Do we need the event for the validated cluster?

manager/pkg/migration/migration_validating.go
```diff
@@ -169,33 +166,173 @@ func IsValidResource(resource string) error {
...
 // It truncates the list to show only the first three clusters.
 func messageClusters(clusters []string) []string {
-	messages := clusters
-	if len(clusters) > 3 {
-		messages = append(clusters[:3], "...")
+	if len(clusters) <= maxClusterMessagesDisplay {
+		return clusters
+	}
+	return append(clusters[:maxClusterMessagesDisplay], "...")
+}
```
yanmxa, line 173: If we already use the event for the managed cluster message, maybe we can remove this function, and just give the count of these invalidated clusters?

yanmxa:
## Suggestion: Simplify Error Message Handling

**File: `manager/pkg/migration/migration_validating.go:32-38`**

The format string constants can be removed and simplified:

```go
// Instead of defining constants like:
const (
	clusterNotFoundMsg = "cluster %s not found in database;"
	clusterNotAvailableMsg = "cluster %s is not available;"
)

// Just use inline strings directly:
if \!ok {
    return fmt.Sprintf("cluster %s not found in database", cluster)
}
if \!isManagedClusterAvailable(mc) {
    return fmt.Sprintf("cluster %s is not available", cluster)  
}
```

This reduces code complexity while maintaining the same functionality for showing different error types in messages.

yanmxa:
## Suggestion: Optimize Database Query Performance  

**File: `manager/pkg/migration/migration_validating.go:302-310`**

The current approach fetches and unmarshals entire JSON payloads:

```go
err := db.Select(columnLeafHubName, columnPayload).
    Where("payload->'metadata'->>'name' IN (?) AND "+columnDeletedAt+" IS NULL", mcm.Spec.IncludedManagedClusters).
    Find(&records).Error
```

**Suggestion**: Consider a two-step approach for better performance:
1. First, get basic cluster-hub mapping (as before)  
2. Only fetch full payload when additional validation is needed

```go
// Step 1: Get cluster-hub mapping efficiently
rows, err := db.Raw(`SELECT leaf_hub_name, cluster_name FROM status.managed_clusters 
    WHERE cluster_name IN (?) AND deleted_at IS NULL`, mcm.Spec.IncludedManagedClusters).Rows()

// Step 2: For clusters needing detailed validation, fetch payload separately
// This avoids JSON unmarshaling overhead for all clusters upfront
```

This maintains the enhanced validation while improving query performance.

yanmxa:
## Issue: String Matching for Control Flow Logic

**File: `manager/pkg/migration/migration_validating.go:195-208`**

```go
// Set condition reason based on error type - use the first error type encountered
if conditionReason == "" {
    if strings.Contains(errorMsg, "not found in database") {
        conditionReason = ConditionReasonClusterNotFound
    } else if strings.Contains(errorMsg, "is already on hub") {
        conditionReason = ConditionReasonClusterConflict
    } else if strings.Contains(errorMsg, "not found in hub") {
        conditionReason = ConditionReasonClusterNotFound
    } else if strings.Contains(errorMsg, "is not available") {
        conditionReason = ConditionReasonResourceInvalid
    } else if strings.Contains(errorMsg, "is hosted") {
        conditionReason = ConditionReasonResourceInvalid
    } else {
        conditionReason = ConditionReasonResourceInvalid
    }
}
```

**Problem**: This code relies on string matching within error messages to determine the appropriate condition reason. This approach is fragile because:

- Changes to error message text will break the logic
- It's prone to false matches if messages contain similar substrings
- The relationship between error types and condition reasons is implicit
- Difficult to maintain and test

The logic couples the user-facing error message format with internal control flow decisions.

PR #2362, opened 2026-04-01
fix: update LastTransitionTime on migration condition reason/message changes

yanmxa:
## Summary

- Add `MigrationCondition` type with `LastUpdateTime` field to track when any condition content changed
- Decouple timeout detection from `LastTransitionTime` by using the previous stage's `LastUpdateTime` as baseline
- Add helper functions (`FindMigrationCondition`, `SetMigrationCondition`, etc.) to replace `meta.SetStatusCondition` for migration conditions

## Details

**Problem:** When a migration stage times out, the `ClusterRegistered` condition's `lastTransitionTime` still shows when the condition was first created (status set to `False/Waiting`), not when the timeout was detected. This is confusing when reading the CR status.

**Root cause:** K8s `meta.SetStatusCondition` only updates `LastTransitionTime` on status changes (e.g., `True` → `False`). When only `Reason` changes (`Waiting` → `Timeout`) while `Status` stays `False`, the time is not updated.

**Solution:** Add `LastUpdateTime` to migration conditions via a custom `MigrationCondition` type that embeds `metav1.Condition`:

```go
type MigrationCondition struct {
    metav1.Condition `json:",inline"`
    LastUpdateTime metav1.Time `json:"lastUpdateTime,omitempty"`
}
```

- `LastTransitionTime`: unchanged, follows K8s convention (updates only on status change)
- `LastUpdateTime`: updates on **any** content change (status, reason, or message)
- Timeout detection uses `latestConditionTime()` which reads `LastUpdateTime` from the previous stage's condition as the baseline

**Example CR after fix:**
```yaml
conditions:
- type: ClusterRegistered
  status: "False"
  reason: Timeout
  lastTransitionTime: "2026-03-31T10:44:54Z"   # when status first became False
  lastUpdateTime: "2026-03-31T10:48:54Z"        # when reason changed to Timeout
```

## References

- Bundle repo update: stolostron/multicluster-global-hub-operator-bundle#905

## Test plan

- [x] Unit tests pass (`go test ./manager/pkg/migration/...`)
- [x] Build passes (`go build ./manager/... ./operator/...`)
- [x] Integration tests compile (`go build ./test/...`)
- [ ] Verify migration timeout scenarios show correct `lastUpdateTime` in CR status

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[ACM-32302]: https://redhat.atlassian.net/browse/ACM-32302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Resolved

- [ACM-32302](https://redhat.atlassian.net/browse/ACM-32302) The time for each migration stage is not clear


operator/api/migration/v1alpha1/condition_helpers.go
```diff
@@ -0,0 +1,112 @@
...
+	if !contentChanged {
+		return false
+	}
+
+	// Status changed: update LastTransitionTime
+	if existing.Status != newCond.Status {
+		if newCond.LastTransitionTime.IsZero() {
+			existing.LastTransitionTime = now
+		} else {
+			existing.LastTransitionTime = newCond.LastTransitionTime
```
clyang82: why do you need to set as `newCond.LastTransitionTime`? how about always set LastTransitionTime to the current time when the status changes?

yanmxa: It should change whenever the status is updated.

clyang82 (approved):
LGTM

PR #2425, opened 2026-05-05
fix: standalone agent e2e Kafka setup failures on release-2.12

birsanv:
## Summary

Fix three root causes that prevented the standalone agent e2e test from running on \`release-2.12\`. All changes are confined to \`test/script/\`.

### 1. Missing \`-n\` namespace flag on \`kubectl wait\` (`event_exporter_kafka.sh`)

\`kubectl wait\` for the standalone agent \`KafkaUser\` was missing the \`-n "$kafka_namespace"\` flag, causing it to look in the \`default\` namespace. This produced a 500-second timeout on every run before falling through to the while-loop fallback. When the fallback also failed, the \`transport-config\` secret was never created in \`open-cluster-management\`, and the test failed with:

\`\`\`
secrets "transport-config" not found  (agent_test.go:45)
\`\`\`

**Fix:** Add \`-n "$kafka_namespace"\` to the \`kubectl wait\` call.

### 2. Race condition: CRDs not Established before \`kubectl apply\` (`e2e_kafka.sh`)

\`kubectl apply -k kafka-cluster\` was running immediately after the Strimzi operator pod became Running, before the Kafka CRDs (\`kafkas\`, \`kafkatopics\`, \`kafkausers\`, \`kafkanodepools\`) were fully registered with the API server. This caused:

\`\`\`
no matches for kind "Kafka" in version "kafka.strimzi.io/v1beta2"
\`\`\`

Also, Kafka 4.0.0 with KRaft mode needs more than the previous 600s to fully populate \`status.listeners\` in a Kind cluster.

**Fix:** Add \`kubectl wait --for=condition=Established\` for all four Strimzi CRDs with a 120s timeout, and increase the Kafka readiness timeout from 600s to 1200s.

### 3. Strimzi `latest` now points to 1.0.0 which removed `v1beta2` (`e2e_kafka.sh`)

Strimzi 1.0.0 (released April 2026) removed the \`v1beta2\` CRD API entirely. Strimzi 0.51.0 also dropped support for Kafka 4.0.x. The test script installed \`https://strimzi.io/install/latest\`, which now resolves to Strimzi 1.0.0 — so \`kafka.strimzi.io/v1beta2\` is never registered on the API server regardless of CRD wait guards.

**Fix:** Pin to **Strimzi 0.50.1**, the last version that supports both \`v1beta2\` and Kafka 4.0.0, by downloading the pinned YAML directly from the GitHub release with the same namespace substitution that the \`strimzi.io/install\` URL used to perform.

## Files Changed

| File | Change |
|------|--------|
| \`test/script/event_exporter_kafka.sh\` | Add \`-n\` flag to \`kubectl wait\` |
| \`test/script/e2e_kafka.sh\` | Pin Strimzi to 0.50.1; add CRD waits; increase Kafka timeout |

## Test Plan

The fix will be validated when \`ci/prow/test-e2e\` runs on this PR — the standalone agent test (\`Standalone Agent [It] should receive the cluster event from the standalone agent\`) should now pass.

Signed-off-by: Valentina Birsan <vbirsan@redhat.com>

### DIFF
--- a/test/script/e2e_kafka.sh
+++ b/test/script/e2e_kafka.sh
@@ -36,12 +36,25 @@ retry "(kubectl get pods -n $kafka_namespace --kubeconfig $KAFKA_KUBECONFIG -l n
 
 echo "Kafka operator is ready"
 
+# Wait for Strimzi CRDs to be fully established before applying Kafka resources.
+# Without this, "kubectl apply -k kafka-cluster" fails with "no matches for kind Kafka"
+# because the CRDs are created but not yet registered with the API server.
+kubectl wait --for=condition=Established crd/kafkas.kafka.strimzi.io \
+  --kubeconfig "$KAFKA_KUBECONFIG" --timeout=120s
+kubectl wait --for=condition=Established crd/kafkatopics.kafka.strimzi.io \
+  --kubeconfig "$KAFKA_KUBECONFIG" --timeout=120s
+kubectl wait --for=condition=Established crd/kafkausers.kafka.strimzi.io \
+  --kubeconfig "$KAFKA_KUBECONFIG" --timeout=120s
+kubectl wait --for=condition=Established crd/kafkanodepools.kafka.strimzi.io \
+  --kubeconfig "$KAFKA_KUBECONFIG" --timeout=120s
+
 node_port_host=$(kubectl config view --minify -o jsonpath='{.clusters[0].cluster.server}' --kubeconfig "$KAFKA_KUBECONFIG" | sed -e 's#^https\?://##' -e 's/:.*//')
 sed -i -e "s;NODE_PORT_HOST;$node_port_host;" "$TEST_DIR"/manifest/kafka/kafka-cluster/kafka-cluster.yaml
+
 # deploy kafka cluster
 kubectl apply -k "$TEST_DIR"/manifest/kafka/kafka-cluster -n "$kafka_namespace" --kubeconfig "$KAFKA_KUBECONFIG"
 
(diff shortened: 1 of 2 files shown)


birsanv:
## Fix: pin Strimzi to 0.50.1 — the real root cause (commit d4c31f16)

### What actually went wrong

All previous fixes were chasing a symptom. The actual root cause:

**Strimzi 1.0.0 was released in April 2026 and removed `kafka.strimzi.io/v1beta2` entirely.**

When the test script ran `https://strimzi.io/install/latest`, it started downloading Strimzi 1.0.0, which only serves the new `v1` CRD API. The test manifests use `apiVersion: kafka.strimzi.io/v1beta2`, so no matter how many retries or cache tricks we applied, the API server simply never registered that group/version — hence "no matches for kind Kafka" on every attempt.

Additionally, Strimzi 0.51.0 (the last `0.x` release) dropped support for Kafka 4.0.0, which is what `kafka-cluster.yaml` requests.

### The fix

Pin to **Strimzi 0.50.1** — the last release that supports both `v1beta2` and Kafka 4.0.0:

```bash
STRIMZI_VERSION="0.50.1"
curl -sL "https://github.com/strimzi/strimzi-kafka-operator/releases/download/${STRIMZI_VERSION}/strimzi-cluster-operator-${STRIMZI_VERSION}.yaml" \
  | sed "s/namespace: myproject/namespace: ${kafka_namespace}/g" \
  | kubectl create -f - -n "$kafka_namespace" --kubeconfig "$KAFKA_KUBECONFIG"
```

The `sed` substitution replicates the namespace injection that `strimzi.io/install/latest?namespace=` used to do automatically.

The unnecessary retry loop and `--cache-dir` complexity from the previous two commits have been cleaned up.

bjoydeep (approved):
lgtm

bjoydeep (approved):
lgtm

PR #2388, opened 2026-04-08
Add DR migration labels and annotations for ZTP resources

ldpliu:
## Summary

- Add `velero.io/restore-status` annotation to ImageClusterInstall resources
- Add `cluster.open-cluster-management.io/backup` label to BareMetalHost resources
- Validate both key and value when setting labels/annotations during migration

### Agent (agent/pkg/spec/migration/migration_from_syncer.go)

**Constants Added:**
- `VeleroRestoreStatusAnnotation = "velero.io/restore-status"`
- `ClusterBackupLabel = "cluster.open-cluster-management.io/backup"`
- `ClusterActivationBackup = "cluster-activation"`

**ImageClusterInstall Processing:**
- Add `velero.io/restore-status: "true"` annotation to prevent image-based-install-operator reconciliation
- Add `velero.io/restore-name: globalhub` label (existing, now validates value)
- Check both key existence AND value correctness

**BareMetalHost Processing:**
- Add `cluster.open-cluster-management.io/backup: cluster-activation` label
- Ensures resources are included in cluster activation backups
- Separated from DataImage case for clearer logic
- Check label value, not just existence

## Why These Changes

**velero.io/restore-status annotation:**
The image-based-install-operator checks for this annotation to determine if a resource was restored by Velero during DR recovery. Without it, the operator may attempt to reconcile ImageClusterInstall resources inappropriately during migration, causing conflicts.

**cluster.open-cluster-management.io/backup label:**
This label marks BareMetalHost resources for inclusion in cluster activation backup workflows, ensuring they're properly captured during DR scenarios.

**Value validation:**
Previously, the code only checked if keys existed. Now it validates values match expected constants, preventing issues where resources have the annotation/label but with incorrect values.

## Test plan

- [ ] Unit tests pass: `make unit-tests-agent`
- [ ] Migration E2E tests verify labels/annotations are set correctly
- [ ] ImageClusterInstall resources have both label and annotation
- [ ] BareMetalHost resources have backup label with correct value
- [ ] Existing migrations continue to work without issues

## Related Issues

https://issues.redhat.com/browse/ACM-32454

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Signed-off-by: daliu@redhat.com

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved migration handling to correctly set/remove backup labels and pause annotations per resource type.
  * Consistently enforce the standardized restore name and add a restore-status annotation to improve restore tracking and visibility.

* **Tests**
[...]

PR #2492, opened 2026-05-30
ACM-32454: Add DR migration labels and annotations for ZTP resources

birsanv:
## Summary

Rebased replacement for #2388. Adds missing DR migration labels and annotations for ZTP resources in \`processResourceByType\`, and validates both key **and** value when setting them.

Supersedes #2388.

## What this changes

During hub migration, the agent's \`MigrationSourceSyncer\` processes ZTP-related resources before they are synced to the target hub. On current \`main\`, that processing is incomplete for two resource types:

| Resource | Current \`main\` behavior | After this PR |
|---|---|---|
| **ImageClusterInstall** | Adds \`velero.io/restore-name\` only if the key is missing (no value check); no restore-status annotation | Sets \`velero.io/restore-name: globalhub\` and \`velero.io/restore-status: "true"\`, correcting wrong values |
| **BareMetalHost** | Combined with DataImage — only removes the Metal3 pause annotation | Removes pause annotation **and** sets \`cluster.open-cluster-management.io/backup: cluster-activation\` |
| **DataImage** | Same combined case as BareMetalHost | Split out; only pause annotation removal (unchanged intent) |

### ImageClusterInstall — \`velero.io/restore-name\` and \`velero.io/restore-status\`

**Context (confirmed with the IBIO team):** The image-based-install-operator uses the \`velero.io/restore-name\` **label** to detect whether an ImageClusterInstall is being restored. When this label is present and the ICI status is empty, IBIO pauses reconciliation and waits ([source](https://github.com/openshift/image-based-install-operator/blob/ba0f5200b1ea07c59f0e059e99d9deebbf4d70c2/controllers/imageclusterinstall_controller.go#L1529)). IBIO separately sets the \`velero.io/restore-status: "true"\` annotation itself during reconcile if it is absent ([source](https://github.com/openshift/image-based-install-operator/blob/ba0f5200b1ea07c59f0e059e99d9deebbf4d70c2/controllers/imageclusterinstall_controller.go#L1538)).

**Why Global Hub must set these during migration:** During hub migration, Global Hub suppresses IBI and BMH resources from being processed through ClusterInstance. This means the SiteConfig operator's default templates (ACM-26864, ACM-26603) never fire for these resources, leaving them without the required DR metadata. Global Hub's \`MigrationSourceSyncer\` must therefore set them directly.

- \`velero.io/restore-name: globalhub\` — **required**: this is the label IBIO checks to pause reconciliation while migration is in progress. Without it, IBIO would begin reconciling the ICI concurrently with Global Hub's migration sync, causing conflicts.
- \`velero.io/restore-status: "true"\` — **pre-set for consistency**: IBIO will set this itself during its first reconcile, but setting it here aligns with the SiteConfig default templates (ACM-26864) and ensures the annotation is present before IBIO runs.

The existing code already set \`velero.io/restore-name: globalhub\`, but only when the label key was absent (no value validation). This PR also validates the value is correct, and adds the missing restore-status annotation.

### BareMetalHost — \`cluster.open-cluster-management.io/backup\`

This label marks resources for inclusion in cluster activation backup workflows. BareMetalHost resources need it so they are properly captured during DR scenarios. Without it, BMH resources are excluded from Velero backup snapshots. The label constants already exist in \`pkg/constants/constants.go\` (\`BackupKey\` / \`BackupActivationValue\`) and are used elsewhere in the backup controller — this PR reuses them rather than redefining local copies.

### Value validation

Previously, label/annotation logic only checked whether a key existed. Resources with the right key but wrong value would pass through unchanged. This PR enforces the expected values, ensuring correct DR configuration even when resources arrive with stale or incorrect metadata.

## Files changed

- \`agent/pkg/spec/migration/migration_from_syncer.go\` — split BareMetalHost/DataImage cases, add backup label and restore-status annotation, value validation
- \`agent/pkg/spec/migration/migration_from_syncer_test.go\` — updated \`TestProcessResourceByType\` expectations

## Why this is still needed

None of the above behavior exists on \`main\` today. The migration syncer still handles these ZTP resource types, but without the DR-specific labels/annotations that downstream operators and backup workflows expect. This gap affects ZTP cluster migration and DR recovery scenarios tracked in ACM-32454.

## Test plan

- [x] \`go test ./agent/pkg/spec/migration/ -run TestProcessResourceByType\`
- [ ] CI: unit, integration, e2e, Konflux

## Related Issues

- https://issues.redhat.com/browse/ACM-32454
- Supersedes #2388

### DIFF
--- a/agent/pkg/spec/migration/migration_from_syncer.go
+++ b/agent/pkg/spec/migration/migration_from_syncer.go
@@ -62,6 +62,9 @@ const (
 	// imageclusterinstall_controller.go#L118
 	VeleroRestoreNameLabel = "velero.io/restore-name"
 
+	// VeleroRestoreStatusAnnotation marks ImageClusterInstall as restored by velero during migration.
+	VeleroRestoreStatusAnnotation = "velero.io/restore-status"
+
 	GlobalHubRestoreName = "globalhub"
 )
 
@@ -442,23 +445,50 @@ func (s *MigrationSourceSyncer) processResourceByType(
 			delete(annotations, HivePauseAnnotation)
 		}
 		resource.SetAnnotations(annotations)
-	case "BareMetalHost", "DataImage":
-		// Remove pause annotation from BareMetalHost and DataImage
+	case "BareMetalHost":
+		// Remove pause annotation from BareMetalHost
+		annotations := resource.GetAnnotations()
+		if annotations != nil {
+			delete(annotations, Metal3PauseAnnotation)
+		}
+		resource.SetAnnotations(annotations)
+
+		// Add cluster backup label with correct value
+		labels := resource.GetLabels()
+		if labels == nil {
+			labels = make(map[string]string)
+		}
+		if labels[constants.BackupKey] != constants.BackupActivationValue {
+			labels[constants.BackupKey] = constants.BackupActivationValue
+		}
+		resource.SetLabels(labels)
+	case "DataImage":
+		// Remove pause annotation from DataImage
 		annotations := resource.GetAnnotations()
 		if annotations != nil {
 			delete(annotations, Metal3PauseAnnotation)
 		}
 		resource.SetAnnotations(annotations)
 	case "ImageClusterInstall":
-		// Add velero restore label if not present
+		// Add velero restore label with correct value
 		labels := resource.GetLabels()
 		if labels == nil {
 			labels = make(map[string]string)
 		}
-		if _, exists := labels[VeleroRestoreNameLabel]; !exists {
+		if labels[VeleroRestoreNameLabel] != GlobalHubRestoreName {
 			labels[VeleroRestoreNameLabel] = GlobalHubRestoreName
 		}
 		resource.SetLabels(labels)
+
+		// Add velero restore status annotation with correct value
+		annotations := resource.GetAnnotations()
+		if annotations == nil {
+			annotations = make(map[string]string)
+		}
+		if annotations[VeleroRestoreStatusAnnotation] != "true" {
+			annotations[VeleroRestoreStatusAnnotation] = "true"
+		}
+		resource.SetAnnotations(annotations)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


birsanv:
**Update: initial PR**

- Rebased onto current `main` with DR migration label/annotation fixes for BareMetalHost and ImageClusterInstall
- Uses shared `constants.BackupKey` / `constants.BackupActivationValue` instead of local duplicates
- Updated `TestProcessResourceByType` — unit tests pass locally

CI should run after `/ok-to-test`.

birsanv:
**Update: add DCO sign-off**

- Amended commit with `Signed-off-by: Valentina Birsan <vbirsan@redhat.com>`
- Fixes missing DCO on `5f38e525`

CI should re-run; DCO check should pass.

giladravid16:
Hey @birsanv, there's no issue from IBIO's perspective to make these changes, but the description about IBIO isn't accurate.
IBIO uses the `velero.io/restore-name` label to determine if the ICI is being restored - you can see it [here](https://github.com/openshift/image-based-install-operator/blob/ba0f5200b1ea07c59f0e059e99d9deebbf4d70c2/controllers/imageclusterinstall_controller.go#L1529).
IBIO sets the `velero.io/restore-status` annotation to `true` during the reconcile if it doesn't exists - you can see it [here](https://github.com/openshift/image-based-install-operator/blob/ba0f5200b1ea07c59f0e059e99d9deebbf4d70c2/controllers/imageclusterinstall_controller.go#L1538).

birsanv:
**Update: correct IBIO description per giladravid16's review**

- Clarified that IBIO uses the \`velero.io/restore-name\` **label** (not the annotation) to detect restore and pause reconciliation
- Clarified that IBIO sets \`velero.io/restore-status: "true"\` itself during reconcile — Global Hub pre-sets it for consistency with SiteConfig templates (ACM-26864), not because IBIO requires it externally
- Added explicit explanation of why Global Hub must set these during migration: SiteConfig default templates (ACM-26864, ACM-26603) never fire because Global Hub suppresses IBI/BMH in ClusterInstance during migration
- Linked to the exact IBIO controller lines referenced by giladravid16

No code changes — description only.

bjoydeep (approved):
lgtm

PR #2657, opened 2026-07-23
ACM-38171: annotate newly imported clusters with HA klusterlet-config

rohit-bharmal:
Fixes: [ACM-38171](https://redhat.atlassian.net/browse/ACM-38171)

## Summary

- When Hub HA is enabled (`hub-role=active`), `HAConfigSyncer` only annotated ManagedClusters that already existed when the HAConfig CloudEvent arrived.
- Adds `HAConfigAnnotator` to watch ManagedCluster create/update and apply `agent.open-cluster-management.io/klusterlet-config` when an `ha-standby-*` KlusterletConfig already exists.
- Refactors shared annotate helper in the syncer and skips the local cluster via `local-cluster=true` or name `local-cluster`.

## Context

QE imported a regional hub into Global Hub, set `deploy-mode=hosted` and `hub-role=active`, then imported a managed cluster (`ks`) onto that regional hub. The new ManagedCluster never received the `klusterlet-config` annotation, because annotation ran only once during the initial HAConfig sync.

## Test plan

- [x] `go test -mod=mod ./agent/pkg/spec/hubha/ ./agent/pkg/spec/`
- [ ] Repro: enable HA on regional hub, then import a managed cluster → annotation is present
- [ ] Confirm `oc get klusterletconfig | grep ha-standby` exists on the regional hub

[ACM-38171]: https://redhat.atlassian.net/browse/ACM-38171?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Automatically applies the correct Hub HA `klusterlet-config` annotation to newly created imported managed clusters.
  * Continues to ignore local hub clusters (by name/role) for HA annotation.
* **Bug Fixes**
  * Improves reliability of HA annotation propagation by updating clusters safely and consistently when HA configuration becomes available.
* **Tests**
  * Added unit tests covering annotation, missing HA configuration, local-cluster skipping, and idempotent behavior.
* **Chores**
  * Minor header and syncer logic cleanup to centralize per-cluster annotation behavior.

PR #2671, opened 2026-07-28
ACM-38171: skip HA klusterlet-config on imported regional hubs

rohit-bharmal:
## Summary

Follow-up to #2657 / ACM-38171. QA found that `ha-config-annotator` also annotated the regional hub ManagedCluster already imported into Global Hub.

- Annotate spoke ManagedClusters only when the agent hub role is **active**
- Skip clusters that must not get the HA annotation: `local-cluster` and hubs with Global Hub `deploy-mode` / `hub-role` labels
- If those labels arrive after create (or a bad annotation is already present), clear the stuck annotation (active or standby)

## Test plan

- [x] `go test ./agent/pkg/spec/hubha/`
- [ ] On Global Hub: imported regional hub ManagedCluster has **no** `agent.open-cluster-management.io/klusterlet-config`
- [ ] On active regional hub: newly imported spoke (e.g. `ks`) **does** get `ha-standby-*`
- [ ] On active regional hub: `local-cluster` stays unannotated

## Summary by CodeRabbit

* **Bug Fixes**
  * HA configuration annotations are now applied only by active hubs.
  * Annotations are removed from local clusters and hubs managed by Global Hub.
  * Incorrect or outdated annotations are cleared when cluster roles or labels change.
  * Standby hubs no longer leave stale HA configuration annotations behind.
* **Tests**
  * Added coverage for active, standby, imported, and Global Hub-managed cluster scenarios.

### DIFF
--- a/agent/pkg/spec/hubha/hubha_config_annotator.go
+++ b/agent/pkg/spec/hubha/hubha_config_annotator.go
@@ -32,6 +32,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
+	"github.com/stolostron/multicluster-global-hub/agent/pkg/configs"
 	"github.com/stolostron/multicluster-global-hub/pkg/constants"
 )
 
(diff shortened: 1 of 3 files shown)


rohit-bharmal:
Update: address CodeRabbit review

- Use unlabeled spoke in `NoOpWhenNotActiveHub` so reconcile reaches `!isActiveHub()`
- Preserve standby hub role setup and no-annotation assertions

@coderabbitai review

birsanv (approved):
lgtm

PR #2689, opened 2026-07-31
chore: backport Konflux pkg touch from release-5.0 to main

birsanv:
## Summary

Backport the `pkg/bundle/version/version.go` comment from #2685 (merged to `release-5.0` only) onto `main` so `main` and `release-5.0` stay aligned after sync PR #2688.

Root cause: Konflux retrigger PRs #2684/#2685 targeted `release-5.0` instead of `main`, breaking main → release-5.0 ffwd.

## Test plan

- [ ] CI passes
- [ ] Merge before or with #2688 so compare shows minimal drift

Related: #2688

Made with Cursor

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Documentation**
  * Added package documentation describing the purpose of the version component.

PR #2690, opened 2026-07-31
fix: harden transport offset migration integration test

birsanv:
## Summary

- `TransportOffsetPersistence` migration test queried **all** `status.transport` rows and expected exactly 3
- The suite-wide manager committer (started in `BeforeSuite`) can insert unrelated offsets in parallel — e.g. spurious `@0` with empty topic from events that fail topic parsing
- Flaked on #2689 integration (`len:4` instead of `len:3`); unrelated to Phase 1-3 or that PR's pkg comment change
- Fix: query only the three expected migrated names (`old-topic-*@*`), same pattern as the multi-partition test below

## Test plan

- [ ] `ci/prow/test-integration` passes
- [ ] Unblocks #2689 once merged (or merge this first)

Related: #2689

Made with Cursor

## Summary by CodeRabbit

* **Tests**
  * Improved migration test validation by checking only the expected migrated transport records.
  * Prevented unrelated records from affecting migration count assertions.

### DIFF
--- a/test/integration/manager/status/transport_offset_test.go
+++ b/test/integration/manager/status/transport_offset_test.go
@@ -96,18 +96,25 @@ var _ = Describe("TransportOffsetPersistence", Ordered, func() {
 			Expect(err).NotTo(HaveOccurred())
 			Expect(oldCount).To(Equal(int64(0)), "old format records should be deleted")
 
-			// Verify new records exist with correct format
-			var newRecords []models.Transport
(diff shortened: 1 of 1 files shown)


birsanv:
**Update: harden transport offset migration integration test**

- Assert only expected `old-topic-*@*` rows, not total table count
- Ignores spurious `@0` offsets from the suite-wide manager committer that flaked #2689 integration

PR #2761, opened 2026-08-18
ACM-34442: tighten managed-hub Kafka ACLs on gh-spec

birsanv:
## Summary

[ACM-34442](https://redhat.atlassian.net/browse/ACM-34442)

- Managed-hub KafkaUsers: Describe+Read only on shared `gh-spec` (no Write)
- Scope consumer-group ACLs to each hub's literal group ID via `GetConsumerGroupID`
- Migration deploying traffic unchanged (`gh-migration` topic + existing ACL reconciler)

## Changes

| File | Change |
|------|--------|
| `strimzi_transporter.go` | Extract `managedHubUserACLs()`; gh-spec Read-only for managed hubs |
| `pkg/utils/utils.go` | `ConsumeGroupReadACL(consumerGroup)` — literal group, not `*` |
| `strimzi_transporter_test.go` | Unit test ACL matrix |
| `transporter_test.go` | Integration test assertions for topic and consumer-group ACLs |

## Test plan

- [x] `go test ./operator/pkg/controllers/transporter/protocol/ -run TestManagedHubUserACLs -count=1`
- [ ] CI integration test `transporter_test.go`
- [ ] Backport to supported release branches after merge

## Related

- ACM-34442

## New Features
- Added automatic Kafka access control for active hub high-availability roles.
- Access is granted or revoked as hubs become active, standby, paused, or deleted.

## Bug Fixes
- Restricted managed hub consumer access to the expected consumer group.
- Removed obsolete wildcard consumer-group permissions.
- Removed specification-topic write access while preserving required read access.
- Improved ACL upgrades while retaining valid migration permissions.

## Tests
- Expanded validation of managed hub and high-availability topic and group permissions.

[ACM-34442]: https://redhat.atlassian.net/browse/ACM-34442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2763, opened 2026-08-19
ACM-34442: tighten managed-hub Kafka ACLs on gh-spec for GH 1.7

birsanv:
## Summary

Backport Phase 4 managed-hub Kafka ACL hardening to `release-2.16` (Global Hub 1.7).

- Remove managed-hub **Write** on `gh-spec` (Describe+Read only)
- Scope consumer-group ACLs to the hub literal instead of `*`
- Drop legacy wildcard/combined ACLs on KafkaUser upgrade reconcile

Related: [ACM-34442](https://issues.redhat.com/browse/ACM-34442), main PR https://github.com/stolostron/multicluster-global-hub/pull/2761

## Test plan

- [ ] `go test ./operator/pkg/controllers/transporter/protocol/...`
- [ ] CI unit/integration jobs on `release-2.16`

[ACM-34442]: https://redhat.atlassian.net/browse/ACM-34442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/operator/pkg/controllers/transporter/protocol/strimzi_transporter.go
+++ b/operator/pkg/controllers/transporter/protocol/strimzi_transporter.go
@@ -342,24 +342,13 @@ func (k *strimziTransporter) isCSVInstalled() (bool, error) {
 	return true, nil
 }
 
-// EnsureUser to reconcile the kafkaUser's setting(authn and authz)
-// set the user can write to status
-func (k *strimziTransporter) EnsureUser(clusterName string) (string, error) {
-	userName := config.GetKafkaUserName(clusterName)
-	clusterTopic := k.getClusterTopic(clusterName)
-
-	authnType := kafkav1beta2.KafkaUserSpecAuthenticationTypeTlsExternal
-	if clusterName == config.GetLocalClusterName() || clusterName == constants.LocalClusterName {
-		authnType = kafkav1beta2.KafkaUserSpecAuthenticationTypeTls
-	}
-
-	simpleACLs := []kafkav1beta2.KafkaUserSpecAuthorizationAclsElem{
-		utils.ConsumeGroupReadACL(),
-		// migration resource into mh: write access to the spec topic is required for cluster migration.
+func managedHubUserACLs(clusterTopic *transport.ClusterTopic, consumerGroupID string) []kafkav1beta2.KafkaUserSpecAuthorizationAclsElem {
+	return []kafkav1beta2.KafkaUserSpecAuthorizationAclsElem{
+		utils.ConsumeGroupReadACL(consumerGroupID),
+		// consume spec bundles from the global hub (read-only; managed hubs must not produce to gh-spec)
 		utils.GetTopicACL(clusterTopic.SpecTopic, []kafkav1beta2.KafkaUserSpecAuthorizationAclsElemOperationsElem{
 			kafkav1beta2.KafkaUserSpecAuthorizationAclsElemOperationsElemDescribe,
 			kafkav1beta2.KafkaUserSpecAuthorizationAclsElemOperationsElemRead,
-			kafkav1beta2.KafkaUserSpecAuthorizationAclsElemOperationsElemWrite,
 		}),
 		// consume migration deploying bundles from the dedicated migration topic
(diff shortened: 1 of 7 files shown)


birsanv:
**Update: fix e2e setup flake (not ACL-related)**

- Failure was in `e2e-setup`: `Failed to initialize hub: global-hub` / `hub oriented command should not running against non-hub cluster`
- `HUB_INIT=false` was passed as a positional arg to `ocm.sh`, so it defaulted to `true` and re-ran `clusteradm init` on global-hub in parallel with other setup steps
- Fixed by exporting env vars before the `bash` call in `test/script/e2e_setup.sh`

birsanv:
**Update: capture local agent logs in e2e_log.sh**

- The Local Agent e2e spec has now failed twice with `multicluster-global-hub-agent` stuck in `CrashLoopBackOff` on the global-hub cluster, while the identical ACL logic passed cleanly on the GH 1.6 backport (#2764)
- `e2e_log.sh` never captured logs for this specific deployment (only hub1/hub2 remote agents), so the actual crash/exit reason has been invisible in both failed runs
- Added a small diagnostic addition to dump `describe`, current, and `--previous` logs for `multicluster-global-hub-agent` on global-hub so the next failure surfaces the real cause

birsanv:
**Update: root cause found — install Policy CRDs on global-hub**

- The diagnostic log capture worked: `multicluster-global-hub-agent` was crash-looping with `FATAL ... failed to create manager: ... no matches for kind "Policy" in version "policy.open-cluster-management.io/v1"`
- This is unrelated to the Kafka ACL changes. `agent/cmd/main.go`'s `initCache` unconditionally watches `Policy` CRs (needed for local-policy support), even for the local agent running directly on global-hub — but `POLICY_INIT` is intentionally `false` for the global-hub↔hub join, so those CRDs were never installed there
- This was previously masked by the `HUB_INIT`/`POLICY_INIT` positional-args bug fixed earlier in this PR: since the env vars weren't actually being passed to `ocm.sh`, `POLICY_INIT` silently defaulted to `true` and installed the CRDs (and full propagator) on global-hub as a side effect. Fixing that bug correctly disabled it, unmasking this missing dependency
- Added a step to install just the Policy CRDs (not the full propagator deployment) on global-hub so the local agent's RESTMapper can resolve the kind

birsanv:
**Update: revert local agent log capture in e2e_log.sh**

- Removed the blanket `kubectl logs` capture for the local agent added earlier — it served its purpose in diagnosing the root cause, but an unfiltered app-log dump isn't worth keeping given these logs land in the publicly-readable Prow/GCS CI artifacts
- The actual fix (installing the Policy CRDs on global-hub) is unaffected and remains in place

birsanv:
**Update: fix path bug in Policy CRD install step**

- The Policy CRD install added earlier was silently skipped: `governance-policy-propagator CRDs not found at /tmp/multicluster-global-hub/test/script/governance-policy-propagator`
- `util.sh`'s `git clone` calls don't `cd` first, so the repo lands in the process's ambient CWD (the repo root, since `make e2e-setup` runs with `cwd=$HOST_DIR`), not under `test/script/`
- Switched to `PROJECT_DIR` (the repo root, already exported by `util.sh`) instead of `CURRENT_DIR` (which pointed at `test/script/`)

birsanv:
**Update: fix unrelated race condition in init\_policy that aborted setup before reaching the Policy CRD fix**

- This failure was **not** a recurrence of the earlier local-agent/Policy CRD issue — the process failed during the parallel hub1/hub2 join step, before e2e_setup.sh ever reached that code
- Root cause: `e2e_setup.sh` runs `init_policy` concurrently once per hub/managed-cluster pair, but all invocations `git clone` governance-policy-propagator/-framework-addon/config-policy-controller into the same shared CWD with no locking — two processes can both see the directory missing and race to clone it, and the loser fails with `destination path already exists`, aborting the whole e2e-setup run
- Added `ensure_git_repo()` (mkdir-based lock) in `util.sh` and used it for all three clone+checkout blocks so concurrent `init_policy` calls serialize instead of racing

PR #2779, opened 2026-08-19
ACM-41773: fix Hub HA standby sync for missing namespaces and local-c…

ab-ghosh:
## Summary
 Managed cluster(s) and their data were not synced successfully to the standby hub. Two independent root causes in the Hub HA standby syncer (`agent/pkg/spec/hubha/hubha_standby_syncer.go`) were fixed:

  1. **Missing namespaces.** Hub HA never syncs `Namespace` objects (`GetHubHAResourcesToSync()` has no Namespace GVK), so per-cluster and application namespaces present on the active hub were absent on the standby. Applying a namespaced resource then failed with `namespaces "X" not found`. A new `ensureNamespace` helper creates the target namespace before the resource is applied. It reads the namespace first (served from the manager cache) and only issues a `Create` when it is genuinely missing, so the hot apply/resync path does not issue a write per resource in steady state.

  2. **local-cluster collision.** The active hub emitted its own `local-cluster` ManagedCluster, which the standby then tried to apply onto its *own* `local-cluster`, getting denied by the OCM webhook (`cluster local-cluster is already local cluster`). `createResource` and `updateResource` now skip global-hub topology ManagedClusters (local-cluster and imported hubs), mirroring the guard that already existed in `cleanupStaleResources` so the apply and cleanup paths treat them the same.

Also corrects the ManagedCluster CRD path in the hubha integration suite (`suite_test.go`) — the wrong filename was silently ignored, so the CRD was never installed and the regression could not be exercised.

## Related issue(s)

Fixes [#ACM-41773](https://redhat.atlassian.net/browse/ACM-41773)

## Tests
 * Added two integration regression tests (envtest) that fail without the fix and pass with it: applying a resource whose
  namespace is absent on the standby, and not overwriting the standby's own `local-cluster`.
 * Full `agent/...` unit suite green (16 packages, 0 failures); full hubha integration + lifecycle suites green

## Summary by CodeRabbit

* **New Features**
  * Standby synchronization now creates missing namespaces before applying namespaced resources.
  * Global Hub topology `ManagedCluster` resources are excluded from standby synchronization and Hub HA emissions.
  * Local cluster configuration is protected from being overwritten during standby synchronization.

* **Bug Fixes**
  * Improved handling of concurrent namespace creation and namespace lookup failures.
  * Prevented stale cleanup and resynchronization from processing excluded topology resources.

* **Tests**
  * Added coverage for namespace creation, local-cluster protection, and topology resource filtering.

### DIFF
--- a/agent/pkg/spec/hubha/hubha_config_annotator.go
+++ b/agent/pkg/spec/hubha/hubha_config_annotator.go
@@ -125,6 +125,14 @@ func shouldSkipHAAnnotation(obj client.Object) bool {
 	return isGlobalHubManagedHub(obj)
 }
 
+// isTopologyManagedCluster reports whether obj is a ManagedCluster that represents the
+// Global Hub topology (local-cluster or an imported hub) rather than a regional cluster
+// synced by Hub HA. These must not be emitted to or applied/deleted on the standby, since
+// they collide with the standby's own topology (e.g. its own local-cluster).
+func isTopologyManagedCluster(gvk schema.GroupVersionKind, obj client.Object) bool {
+	return gvk.Group == clusterv1.GroupName && gvk.Kind == "ManagedCluster" && shouldSkipHAAnnotation(obj)
+}
+
 func hasHAKlusterletConfigAnnotation(obj client.Object) bool {
 	return obj.GetAnnotations()[klusterletConfigAnnotation] != ""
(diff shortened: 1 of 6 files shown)


birsanv (approved):
lgtm

PR #2879, opened 2026-09-03
ACM-42804: skip hub-role annotation for regional hubs

rohit-bharmal:
Skip copying hub-role label to addon annotations for regional hubs to prevent addon recreation.

Regional hubs are identified by `addon.open-cluster-management.io/on-multicluster-hub=true` annotation.

Completes the fix from PR #2833.

Jira: https://redhat.atlassian.net/browse/ACM-42804

## Summary by CodeRabbit

- **Bug Fixes**
  - Regional hubs imported into the multicluster environment no longer receive hub-role annotations on their managed add-ons.
  - Managed clusters continue to retain hub-role annotations when applicable.
  - Add-ons no longer receive an unnecessary hub-role annotation when no hub-role label is present.
  - Hub-role annotation behavior is now consistent across regular managed clusters, hosted regional hubs, and highly available managed hubs.


mihirlele:
@coderabbitai review

